PR #9, opened 2019-01-22
Revert "See if this fixes travis failures"

karawoo:
New synapse python client is released, so this shouldn't be needed. This reverts commit ce2b8bb0bb7ee9c773a4ca426cf63d061ed04f46.


karawoo:
This was a different issue described here: https://sagebionetworks.jira.com/browse/SYNR-1375

The remotes one is still present but the changes in this PR shouldn't affect it.

PR #16, opened 2019-01-25
Add functions to return valid annotations

### LINKED ISSUE #5: Add function to show *valid* annotations
Could be helpful to see which valid annotation keys and values are being used.

karawoo:
I refactored the code a little so it can return either valid or invalid annotations depending on what the user asks for. To view invalid annotation keys/values, use `check_annotation_keys()` and `check_annotation_values()` as before; to view valid ones, use `valid_annotation_keys()` and `valid_annotation_values()`. Closes #5.


karawoo:
The failed build was just due to the tests not passing because of the values from Synapse being in a different order than the test expected. I agree that sorting is an unsatisfying way to deal with it, I'm not sure off the top of my head whether there's another good way around it though. I'll merge for now, we can always improve the tests later.

PR #20, opened 2019-01-31
Update README

karawoo:
Shows the new functionality from #19 and also includes output so people can see how it actually looks to run these functions.


karawoo:
the one failing build is due to travis having trouble installing R, nothing related to this PR

PR #31, opened 2019-03-08
Make check_type() omit NAs and duplicates

### LINKED ISSUE #30: check_type() marks NAs as invalid and includes duplicates
When checking annotations in a file view, `NA` values get returned as invalid when the annotation type is being checked. This came up when checking annotations in AMP-AD: the file view contains the annotation PMI, which according to our annotation dictionary should be a string. All of the files that use this annotation have it as a number; as a result the data frame column is numeric. But many files don't have this annotation at all. `check_annotation_values()` returns all of the values as invalid. The correct behavior would be to return only the (unique) values and not the `NA`s.

karawoo:
Fixes #30


karawoo:
No worries!

PR #35, opened 2019-04-26
Whitelist annotations

### LINKED ISSUE #32: Allow whitelisting certain annotation keys/values
I'm thinking it would be good to add "whitelisting" options for annotation keys or values that are technically invalid but that we don't want to check. For example, `grant` is an annotation key, but sometimes we don't keep up with adding all new grant numbers, and it would be nice to not object if someone puts in their own grant number.

karawoo:
Closes #32.

You can now whitelist certain annotation keys/values while checking. Previously this was possible by providing subsetted data to the `annotations` argument, but the changes in this PR make it easier to provide a few exceptions to our annotations dictionary without having to edit the annotations data locally.

Sample usage:

```r
# Whitelist a key
check_annotation_keys(dat, whitelist_keys = "assay")

# Whitelist all values for the assay key
check_annotation_values(dat, whitelist_keys = "assay")

# Whitelist the value foo from the assay key
check_annotation_values(dat, whitelist_values = list(assay = c("foo")))
```


karawoo:
The build is failing in this and #36 because I made this pull request from my fork, and the encrypted Travis environment variables aren't available from forks. I did check these both before creating the PR and the builds succeed locally, but let me know if you want me to close this PR and create a new one from a branch within the `Sage-Bionetworks` copy of the repo.

PR #39, opened 2019-07-02
Implement custom conditions for reporting check results

### LINKED ISSUE #37: Consistent structure for reporting results
To support https://github.com/Sage-Bionetworks/dccvalidator-app/issues/15, create a more consistent structure for reporting results. 

See section on custom signal classes in advanced R: http://adv-r.had.co.nz/Exceptions-Debugging.html#condition-handling, though we'd want a bit more information than what is in those examples.

Should include:

- Statement of the correct behavior
- Statement of the observed behavior and how it relates to the correct behavior (e.g. some specimens are missing from one of the metadata files)
- Required vs. optional (vs. informational? KNB has this, I'm not certain we need it). This may vary for different use cases, so either should be implemented as an option to the check function, or should be handled by the shiny app.

karawoo:
This PR implements custom condition classes for reporting the results of `check_*()` functions. This provides a consistent structure for reporting results (and closes #37): each `check_*()` function returns a condition object of class `"check_pass"`, `"check_warn"`, or `"check_fail"`. These custom conditions are helpful because they can contain data in addition to signaling the status of the check result. Right now the information contained in the condition is: the message to display (i.e. the results of the check), a statement of the correct behavior (what the function was checking for), and---in the case where there is invalid data---the data values that were invalid. This data is what was returned by the functions previously, so it should be relatively easy to update code. If you previously used the resulting data like so:

```r
foo <- check_annotation_keys("bar")
do_some_thing(foo)
```

now you would replace `do_some_thing(foo)` with `do_some_thing(foo$data)`.

A more complete example of the new functionality and resulting object is below:

``` r
library("synapser")
synLogin()
annots <- get_synapse_annotations()

## Get a file to check
my_file <- synGet("syn11931757", downloadFile = FALSE)
my_file$annotations
#> {
#>   "annotation_a": [
#>     2
#>   ],
#>   "annotation_b": [
#>     "test"
#>   ]
#> }

## Check the file:
result <- check_annotation_keys(my_file, annots)

## View condition object:
result
#> <error>
#> message: Some annotation keys are invalid
#> class:   `check_fail`

## View structure:
str(result)
#> List of 5
#>  $ message : chr "Some annotation keys are invalid"
#>  $ trace   : NULL
#>  $ parent  : NULL
#>  $ behavior: chr "All annotation keys should conform to the vocabulary"
#>  $ data    : chr [1:2] "annotation_b" "annotation_a"
#>  - attr(*, "class")= chr [1:4] "check_fail" "rlang_error" "error" "condition"

## Access data:
result$data
#> [1] "annotation_b" "annotation_a"
```

Note that the check functions do not _throw_ errors/warnings/messages; they create objects whose classes inherit from error/warning/message. You could, however, choose to signal an error/warning/message like so:

```r
rlang::cnd_signal(result)
#> Error: Some annotation keys are invalid
```

The purpose of this is to create more consistency and enable easier reporting in the dccvalidator app (https://github.com/Sage-Bionetworks/dccvalidator-app/issues/15).

@milen-sage, @kdaily, @kelshmo, @bememea, and anyone else who a) is using this package and b) has made it to the end of this long explanation, will you let me know if you have any concerns about this change? The package is still in an [experimental](https://www.tidyverse.org/lifecycle/#experimental) phase, but I also don't want to majorly disrupt your work.


karawoo:
Confirmed this won't mess things up for @milen-sage or @bememea, will have to talk to others after they're back in the office.

kelshmo:
I don't have any concerns - thanks!

PR #47, opened 2019-09-06
Add function to test for empty columns

karawoo:
This function returns a warning (by default) or a failure (if `strict = TRUE`) if any of the columns are completely empty (by default "empty" means `NA` or `""` but this can be overridden). This is to support https://github.com/Sage-Bionetworks/dccvalidator-app/issues/43

Note that a column may be _partially_ empty and not raise this warning. We could check for any missing data at all, but I suspect that would not be very useful as it's very common for some rows to have data missing because it's not relevant for that observation.


tests/testthat/test-check-cols-empty.R
```diff
@@ -0,0 +1,38 @@
...
+  dat3 <- data.frame(x = 1, y = NA, z = "")
+  res1 <- check_cols_empty(dat1)
+  res2 <- check_cols_empty(dat2)
+  res3 <- check_cols_empty(dat3)
+  expect_true(inherits(res1, "check_warn"))
+  expect_true(inherits(res2, "check_warn"))
+  expect_true(inherits(res3, "check_warn"))
+  expect_equal(res1$data, "x")
+  expect_equal(res2$data, "x")
+  expect_equal(res3$data, c("y", "z"))
```
kelshmo: nice!

PR #62, opened 2019-09-13
Use addResourcePath() to find css and js resources

karawoo:
I believe this is the recommended way to load external resources (custom CSS and the javascript we use to check authentication) with Shiny, and should allow an app to find these resources as long as dccvalidator is installed, even if the app is stored in a different location.

### DIFF
--- a/R/app-ui.R
+++ b/R/app-ui.R
@@ -132,12 +132,15 @@ app_ui <- function(request) {
 
 #' @import shiny
 golem_add_external_resources <- function() {
+
+  addResourcePath(
+    "www", system.file("app/www", package = "dccvalidator")
+  )
+
   tags$head(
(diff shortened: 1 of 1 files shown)


karawoo:
Just to add more detail about what this is doing: before this PR, the app uses relative paths to find the javascript and CSS files from where the app is run (in our current deployment for AMP-AD, we clone this repo and run the `app.R` file at the top level). This PR changes things so that you could have a copy of the app running from a location that doesn't include the whole git repo, as long as the dccvalidator package is installed somewhere. `addResourcePath()` is a shiny function that adds a directory of resources to the web server; in this case, it looks up the location of the files wherever the dccvalidator package is installed, and makes them available to the app.

PR #66, opened 2019-09-20
Fix style issues

karawoo:
I believe this fixes all of the style issues caught by the linter in #64.

### DIFF
--- a/R/app-server.R
+++ b/R/app-server.R
@@ -8,7 +8,9 @@ app_server <- function(input, output, session) {
     showModal(
       modalDialog(
         title = "Not logged in",
+        # nolint start
         HTML("You must log in to <a href=\"https://www.synapse.org/\">Synapse</a> and be a member of the AMP-AD Consortium team to use this application. Please log in, and then refresh this page. If you are not a member of the AMP-AD Consortium team, you can request to be added at <a href=\"https://www.synapse.org/#!Team:3320424\">https://www.synapse.org/#!Team:3320424</a>.")
+        # nolint end
       )
     )
   })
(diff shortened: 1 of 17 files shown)


tthyer:
This looks good to me, for what that's worth. You might want to add a pre-commit hook for this.

karawoo:
Thanks @tthyer! Yeah I'm planning to look into adding a pre-commit hook as well.

R/app-server.R
```diff
@@ -8,7 +8,9 @@ app_server <- function(input, output, session) {
     showModal(
       modalDialog(
         title = "Not logged in",
+        # nolint start
```
kdaily: Only a question - why use the `#nolint start/end` for a single line? I noted in other places you have a long line and did `#nolint` at the end. I see other places that use the `start/end` method for blocks of code, which makes sense.

karawoo: Because I made a mistake 😅 . I meant to use `# nolint` for single lines, and a block for anything that was more than one line.

karawoo:
Thanks @kdaily! To explain my choices of linters:

* `object_length_linter(40)` -- this checks that the lengths of objects/functions aren't too long. Because some of my functions are S3 methods, they can be longer than recommended by default (e.g. `check_annotation_values.CsvFileTable`).
* `object_usage_linter = NULL` -- I disabled it for two reasons: 1) it complains about any function that's used without at `::` prefix. Such functions appear in test files, and in the app functions where I have imported all of shiny to avoid having to write `shiny::` a million times. 2) If an out of date version of the package is installed, it can create spurious errors about undefined functions in the up-to-date version of the package. So correct, exported functions cause errors because they didn't exist in whatever previous version of the package was installed.
* `object_name_linter = NULL` -- this complains about S3 methods and functions from other packages not being in snake_case. I can't do anything about functions from other packages using camelCase, nor is there a way to avoid using periods in S3 method definition, so I disabled it.
* `commented_code_linter = NULL` -- I disabled it because it's not that helpful and complains about double `##` for comments, which are common in the codebase because ESS uses them by default.

kdaily:
All sensible to disable - thanks for taking the time to describe!

PR #68, opened 2019-09-20
Warn about empty columns in metadata files and manifest

### LINKED ISSUE #49: Note empty values and x,y df
These are suggestions more than issues: 

- Note empty columns with a warning in place of a success message. For example, Successes: "All biospecimen columns present" might also be associated with Warning: "Missing organ, tissue,BrodmannArea values". 

- Be descriptive with x and y noting "missing from biospecimen file" and "missing from manifest"
Screen Shot 2019-09-03 at 3 34 58 PM

thanks for this tool!

karawoo:
Closes #49. Checks each uploaded file for empty columns and reports them as warnings.

### DIFF
--- a/R/app-server.R
+++ b/R/app-server.R
@@ -204,6 +204,37 @@ app_server <- function(input, output, session) {
       check_specimen_ids_dup(biosp())
     })
 
+    # Empty columns produce warnings -----------------------------------------------
+    empty_cols_manifest <- reactive({
+      check_cols_empty(
+        manifest(),
+        success_msg = "No columns are empty in the manifest",
+        fail_msg = "Some columns are empty in the manifest"
+      )
+    })
+    empty_cols_indiv <- reactive({
+      check_cols_empty(
+        indiv(),
+        success_msg = "No columns are empty in the individual metadata",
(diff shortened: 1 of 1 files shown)


R/app-server.R
```diff
@@ -187,13 +195,46 @@ app_server <- function(input, output, session) {
...
     duplicate_specimen_ids <- reactive({
       check_specimen_ids_dup(biosp())
     })
 
+    # Empty columns produce warnings -------------------------------------------
+    empty_cols_manifest <- reactive({
+      check_cols_empty(
+        manifest(),
+        success_msg = "No columns are empty in the manifest",
+        fail_msg = "Some columns are empty in the manifest"
```
kelshmo: How is a warning message coded?

kelshmo: Looks like if the argument `strict = FALSE` is in `check_cols_empty`, a warning is thrown instead of an error.

karawoo: Correct: by default `check_cols_empty()` creates a warning object for empty columns, but if `strict = TRUE` then it creates an error object. The `fail_msg` is the same in either case.

Issue #56, opened 2019-09-13
Add automated testing


karawoo:
Need to figure out if/how [shinytest](https://rstudio.github.io/shinytest/index.html) will work with the session token

karawoo:
High priority, but likely difficult...

karawoo:
See here for detailed information on testing applications in a package: https://rstudio.github.io/shinytest/articles/package.html

karawoo:
I spent some time with this and shinytest doesn't currently work because of the way we get the user's login token cookie. When the test script runs, it can't authenticate. We need to figure out a workaround. I also had issues figuring out the correct directory structure within `tests/`.

I learned from this [blog post](https://nadirsidi.github.io/Shinytest/) that there's a global `"shiny.testmode"` option which might be helpful. Perhaps we could wrap all the server logic in a module, and call it within the `observeEvent(input$cookie, ...)` when not in test mode, and in test mode use other credentials.

karawoo:
Adding notes to myself: @Aryllen has done a ton of work on the `shiny-testing` branch to try to get this working. On travis, and when I run `devtools::test()` (or `devtools::test(filter = "shiny")`) we see an error:

```
── 1. Error: Shiny app works (@test-shiny-app.R#8)  ────────────────────────────
No test scripts found in apps/tests
1: expect_pass(testApp("apps", compareImages = FALSE)) at testthat/test-shiny-app.R:8
2: testApp("apps", compareImages = FALSE)
3: stop("No test scripts found in ", testsDir)
```

`devtools::check()` does not throw the error, presumably because the test includes `skip_on_cran()` and `devtools::check()` runs with `--as-cran`? If I comment out `skip_on_cran()` I get the same error as above.

It looks like shinytests looks for test files directly underneath whatever is given as the test directory; it doesn't look recursively to find test scripts further down: https://github.com/rstudio/shinytest/blob/8a72535b466f5ce5f521e1b06ff66df6dd07d9df/R/test-app.R#L180-L182. So that might explain the error we are seeing.

kelshmo:
Adding notes from my conversation with Kara: Existing testing checks behavior of individual functions, while shinytest checks the behavior/functionality of the app is preserved. @Aryllen has done significant work on this.

Aryllen:
[...]

PR #73, opened 2019-09-25
Refactor team lookup and storing data

karawoo:
This extracts some of the code in the app into separate functions (`get_user_teams()` and `create_folder()`) and a shiny module (`check_team_membership()`). The app currently. still hard codes the location of the folder to store data and team to check membership of, but this will make it easier in the future for these details to be configurable as they're now function arguments.


R/get-user-teams.R
```diff
@@ -0,0 +1,25 @@
...
+#' @param user Synapser user object (e.g. output from
+#'   [synapser::synGetUserProfile()])
+#' @return Character vector of team IDs the user belongs to
+#' @examples
+#' \dontrun{
+#' get_user_teams()
+#' get_user_teams("karawoo")
+#' }
+get_user_teams <- function(user) {
+  user_teams <- synapser::synRestGET(
```
thomasyu888: Does synapser have the paginated rest get convenience function?

tests/testthat/test-get-user-teams.R
```diff
@@ -0,0 +1,10 @@
+context("test-get-user-teams.R")
+
+library("synapser")
+if (on_travis()) syn_travis_login() else synLogin()
```
thomasyu888: I see that you have a synapse login in Travis.  Just out of curiosity, what would the code be to mock the synapse store function

karawoo: I have no idea honestly. I've done mocking before for simple HTTP responses, but I'm not sure what it would take to mock the synapser/PythonEmbedInR objects.

R/get-user-teams.R
karawoo: Not sure which one you mean?

thomasyu888: The python client has a hidden function _GET_paginated which does the pagination of rest calls such that we don't have to set the limit really high or deal with getting all the results of a rest call.  Was just wondering if synapser had this hidden function.

karawoo: Ahh I didn't know about that. No, synapser doesn't have any of the hidden functions from the python client 😞

karawoo: There's been some jira discussion about this but no conclusions as far as I know https://sagebionetworks.jira.com/browse/SYNPY-996

karawoo:
After chatting with @kelshmo I rewrote the module as a regular function. It probably doesn't need to be a module (even though it creates shiny app components) since it's not dealing with any inputs/outputs and therefore doesn't have the namespace issues that shiny modules solve. This allowed me to half-test the behavior: if the user is in the team, the function should return `NULL`. I can't currently test the case where the user isn't in the team unless I create a whole app for testing and run shinytest on it. That will be part of #56.

I made a small tweak to the arguments of `check_team_membership()` so it can take the user object directly instead of looking up the user's teams and passing those in.

R/get-user-teams.R
```diff
@@ -0,0 +1,25 @@
...
+#' current logged in user. You must be logged in to Synapse to use this
+#' function.
+#'
+#' @param user Synapser user object (e.g. output from
+#'   [synapser::synGetUserProfile()])
+#' @return Character vector of team IDs the user belongs to
+#' @examples
+#' \dontrun{
+#' get_user_teams()
+#' get_user_teams("karawoo")
```
thomasyu888: This example would fail.  You would need to pass in `synGetUserProfile("karawoo")`. Looks good otherwise!

karawoo: Good catch, I changed the behavior at one point and forgot to update the example. Fixed now. Thanks!

PR #74, opened 2019-09-26
Allow annotation values that are coercible to strings

### LINKED ISSUE #70: Try to coerce values before checking type
Some of our annotation values have weird requirements about types. For example, `readLength` is expected to be a string, but most data contributors provide it as a number. This is not really a problem and shouldn't fail validation as long as the value is coercible to the correct type. It will be highly confusing to users to see perfectly valid read lengths marked as invalid. We should attempt to coerce the values and only report invalid types if they're not coercible.

karawoo:
I'm opening this as a draft because I'm not yet sure if I think it's the right thing to do. This would close #70.

The current issue is that some reasonable annotation values are rejected as invalid because they are of the wrong type. For example, `readLength` is defined as a [string](https://github.com/Sage-Bionetworks/synapseAnnotations/blob/c4c0f4f25c48271877cd596b6356891edbacf1b8/synapseAnnotations/data/ngs.json#L97-L101), but most people will provide it as a number. It will be confusing for the contributor to see that their numbers are invalid, and on our end it's probably not important to prevent numeric read lengths, since they can easily be coerced to strings.

This PR updates the `check_type()` function so that if a numeric/integer/boolean value is provided for a string annotation, it is accepted as valid, since all of these types should be easily coercible. `check_type()` only runs for annotations that _don't_ have an enumerated list of allowable values, so this won't affect anything that has a specific list of allowable values.

Is this too permissive? An alternative approach is just to whitelist specific keys you don't want to enforce (this is already supported in `check_annotation_keys()` and `check_annotation_values()`).


kdaily:
The solution I thought of was to require that when individuals submit a table (in my case, CSV) for validation that CSV conventions are used to indicate the column formats - e.g., all strings are quoted. If we are reading from a File View, this could also be the case (set column schema to the appropriate type that matches the most permissive/castable type that is defined in the schema). This basically makes the data reading schema-aware (which can be done with a CSV or file view).

For example, if a key has `boolean` and `string` allowed in it, then when reading the data all values for that key are converted to `string`. If it only had `boolean` in it, then it would be left as is, and providing a `string` would fail (unless your string was a pre-defined boolean string, like the string `'True'`).

One thing that @cw-dvr8 and I came across was that some schema definitions use `anyOf` instead of `enum` making this determination difficult - you have to inspect all of the fields of `anyOf` to figure out what are the potential types, as using `anyOf` doesn't require that you also set `types` or `enum`s.

Or, that might just make things too darn confusing.

karawoo:
Am I understanding correctly @kdaily that with either approach (yours or the one in this PR), the end result is that the data the user provides stays however they entered it and is treated as valid, even if for example their readLength column is numeric? That is, you're not changing the data that gets uploaded, just converting the columns for the purpose of validation?

> One thing that @cw-dvr8 and I came across was that some schema definitions use anyOf instead of enum making this determination difficult - you have to inspect all of the fields of anyOf to figure out what are the potential types, as using anyOf doesn't require that you also set types or enums.

The code in this draft PR is still using the current synapseAnnotations on master, not the generated JSON Schemas, so this shouldn't be an issue just yet as values can only have one type. However the use of `anyOf` in those schemas should also be resolved when we make the switch over to managing schemas with schema.org and generating JSON Schemas from them. When I created JSON Schemas from our annotations, I used `anyOf` so that we could retain the source and descriptions along with the values, but that's not really necessary as long as we retain the sources in whatever data we generate the schemas from.

karawoo:
@larssono thinks this is a good idea

karawoo:
Adding context to my comment above: we agreed that it was reasonable to be this permissive and if we need to exert finer control on which keys/values are accepted, then we can change our approach in the future to whitelist those specific keys/values.

This still leaves the question of whether we do what's in this PR (treat anything as coercible to string), or whether we use @kdaily's approach of actually coercing the values when the data is read in. I feel like this PR's approach is a little simpler but I'm open to doing the other way.

kdaily:
:+1: we can reevaluate in the context of JSON schema later. Simple is good!

karawoo:
Thanks both!

PR #78, opened 2019-09-26
Changed manifest requirements to template

### LINKED ISSUE #76: Require additional values for manifest
In our last sys bio DCC meeting I believe we agreed to require additional annotations on the manifest that don't appear in the metadata templates (things like consortium, study, grant, fileFormat). I've begun a [template manifest](https://www.synapse.org/#!Synapse:syn20820080) for AMP-AD. `check_cols_manifest()` should be updated to look up a template the way the other ones do.

Aryllen:
Fixes #76 
Changed the required manifest annotations to be in the form of a template. Added manifest template id.

### DIFF
--- a/R/check-col-names.R
+++ b/R/check-col-names.R
@@ -28,6 +28,8 @@ check_col_names <- function(data, template, success_msg = NULL, fail_msg = NULL,
 #' @rdname check_col_names
 #' @examples
 #' \dontrun{
+#' library("synapser")
+#' synLogin()
 #' a <- data.frame(path = "/path/file.txt", parent = "syn123", assay = "rnaSeq")
 #' check_cols_manifest(a)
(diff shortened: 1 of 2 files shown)


Aryllen:
Sure. Will work on those items.

R/check-col-names.R
```diff
@@ -27,20 +27,22 @@ check_col_names <- function(data, template, success_msg = NULL, fail_msg = NULL,
 #' @export
 #' @rdname check_col_names
 #' @examples
-#'
+#' \dontrun{
```
karawoo:
```suggestion
#' \dontrun{
#' library("synapser")
#' synLogin()
```

I'd suggest adding these two lines so that someone could directly run the example, otherwise looks great!

PR #81, opened 2019-09-27
Ensure check_value extracts annotation values as a vector even when input is a tibble

### LINKED ISSUE #79: check_value() doesn't work when given a tibble
If we pass annotations as a tibble to `check_value()` AND there is more than one value enumerated for the given key, the results are incorrect:

``` r
library("dccvalidator")
#> Loading required package: shinyBS
library("tibble")

annots <- tribble(
  ~key, ~value, ~columnType,
  "fileFormat", "txt", "STRING",
  "fileFormat", "csv", "STRING",
  "species", "Human", "STRING"
)

dccvalidator:::check_value(
  values = "Human",
  key = "species",
  annotations = annots
)
#> character(0)

dccvalidator:::check_value(
  values = "txt",
  key = "fileFormat",
  annotations = annots
)
#> [1] "txt"
```

<sup>Created on 2019-09-26 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

`"txt"` should be a correct value. 

Problem is here:

https://github.com/Sage-Bionetworks/dccvalidator/blob/3603c40e0b5ec967decb742646c76948f0634d84/R/check-annotation-values.R#L226

We should make sure that the result of this line is a vector, not a data frame/tibble. For some reason the [`%in%` step here](https://github.com/Sage-Bionetworks/dccvalidator/blob/3603c40e0b5ec967decb742646c76948f0634d84/R/check-annotation-values.R#L244) works when there is only one value, but not when there are more than one.

``` r

library("tibble")
dat <- tibble(foo = c("a", "b"))
"a" %in% dat
#> [1] FALSE
"a" %in% dat[1, ]
#> [1] TRUE
```

<sup>Created on 2019-09-26 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

karawoo:
Fixes #79

### DIFF
--- a/R/check-annotation-values.R
+++ b/R/check-annotation-values.R
@@ -223,7 +223,7 @@ check_value <- function(values, key, annotations, whitelist_keys = NULL,
   if (!key %in% annotations$key) {
     return(NULL)
   }
-  annot_values <- annotations[annotations$key == key, "value"]
+  annot_values <- annotations[annotations$key == key, ]$value
   ## If key is being whitelisted, treat all values as valid
   if (key %in% whitelist_keys) {
     if (isTRUE(return_valid)) {
--- a/tests/testthat/test-check-annotation-values.R
+++ b/tests/testthat/test-check-annotation-values.R
@@ -131,6 +131,16 @@ test_that("check_value checks column type when no enumerated values", {
   )
 })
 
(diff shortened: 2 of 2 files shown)


karawoo:
Thank you both!

PR #82, opened 2019-09-27
Skip tests that require authentication when on fork

karawoo:
* Adds a function `skip_on_fork()` that we can use to skip running certain tests on Travis builds from forks. Currently PR builds from forks fail because the encrypted environment variables are not available to forks, so we'll skip the tests that need those variables, and PR builds should then generally pass unless there are other issues. 

  This is obviously not 100% ideal as there could be issues that those tests would have caught that will then make it through to master. But that is already the case now anyway, and this makes it easier to catch other issues that cause the build to fail outside of these tests. The long-term better solution would be to mock all content from synapse, but figuring out how to mock PythonEmbedInR objects is a bigger project for another time.
* Updates a number of tests to use local data rather than relying on getting data from Synapse, so fewer tests will need to be skipped. 
* Removes some calls to `skip_on_cran()`; realistically, this package will never go to CRAN since it depends on synapser.


karawoo:
I updated this to simplify the decision about when to skip: rather than detecting the exact conditions the test is running under ("is this a) on Travis, b) a pull request, and c) from a fork?"), I have set the tests that use Synapse to simply skip if the user is not logged in after `attempt_login()` has been called.

tests/testthat/test-utils.R
```diff
@@ -1,18 +1,25 @@
...
-  }
+  skip_if_not(on_travis())
+
+  ## In this one we do need to ensure it only runs for builds on upstream repo,
+  ## not forks
+  owner <- gsub(
+    "(^[^/]+)(.+)", "\\1",
+    Sys.getenv("TRAVIS_PULL_REQUEST_SLUG")
+  )
+  skip_if_not(owner == "Sage-Bionetworks", "Testing on upstream repo")
```
thomasyu888: Only line with explanation of why test is skipped.  Did you want the others have some context like:
```skip_if_not(logged_in(), "Not Logged In")```?  

Not a big deal. Looks good!

karawoo: Mentioned this in slack but I'll put it here too for posterity: there’s some default text that will appear, so they only need a message if you want to override that. skip_if_not(logged_in()) will show "logged_in() is not TRUE" when it’s skipped — I thought that was sufficient in most cases

PR #84, opened 2019-10-01
Allow checking membership in multiple teams

### LINKED ISSUE #83: Allow app to check against multiple teams
For PEC, the user may be a part of either Team A or Team B. 

Modify `check_team_membership()` to query against more than one team ID.

karawoo:
Fixes #83 

I split part of the `check_team_membership()` function into its own function, `report_missing_membership()`. `check_team_membership()` now just checks whether the user is in the right team(s); if not, the modal dialog is created by `report_missing_membership()`.


R/app-reporting.R
```diff
@@ -82,3 +82,45 @@ show_details.list <- function(x) {
...
+#' the required teams, creates a modal dialog indicating which teams they need
+#' to belong to and how to request access.
+#'
+#' @param result Output from [check_team_membership()]
+report_missing_membership <- function(result) {
+  UseMethod("report_missing_membership", result)
+}
+
+#' @rdname report_missing_membership
+report_missing_membership.default <- function(result) {
```
kelshmo: I'm not yet very familiar with S3 objects - am I understanding correctly that `check_team_membership()` returns an object with a class of `check_pass` or `check_fail`, `report_missing_membership()` sticks to a default output unless it sees a `check_fail` class? I'm curious to learn in what cases you choose to implement this object type.

R/check-team-membership.R
```diff
@@ -1,22 +1,29 @@
...
-    showModal(
-      modalDialog(
-        title = glue::glue("Not in {team_name} team"),
-        # nolint start
-        HTML(
-          glue::glue("You must be a member of the {team_name} team on Synapse to use this tool. If you are not a member of the {team_name} team, you can request to be added at <a href=\"https://www.synapse.org/#!Team:{team}\">https://www.synapse.org/#!Team:{team}</a>.")
-        )
-        # nolint end
-      )
+  team_names <- glue::glue_collapse(
```
kelshmo: O a new tidy function!

R/app-reporting.R
karawoo: Yup that's right. Now that I'm looking at it again I think this might be overkill, so I pushed an update. A better example of a time when it's a good idea to use S3 is the `check_annotation_keys()` function which does slightly different things for data frames, Synapse files, and file views. If you don't write them as separate methods, then you end up with really long and complex if statements to ensure that the right thing happens for the right type. Another time that S3 methods come in very handy is when you want to define a new method for a commonly used function that you don't control (like writing a `print()` method for a new object type you've created).

PR #88, opened 2019-10-02
Prevent crashes when user isn't in correct team

### LINKED ISSUE #86: App behavior when not signed in to Synapse or not part of a team
1. If user is not signed in to Synapse, user can still browse and upload a file without any results. 
2. If user is signed in to Synapse but not a part of a team, the "Not in team" message is displayed grayed out and server session is disconnected. The link to be added to a team is displayed and not clickable.

I think use case 1 is fine as one, or alternatively we could find a way that the message is displayed and it is not possible to click out of it to upload files.

For use case 2, since the link is not clickable, does it make sense to add a URL that is more easily searchable? Alternatively, we could devise another path to a resolution.

karawoo:
Fixes #86. Alters the app to:
* Only create a folder for saving stuff if the user is in the correct team
* Disable inputs on startup, and enable them only once the app has established that the user is a member of the team. This will prevent errors if the user is logged in but not a member of the team by preventing users from uploading data that would then fail to upload due to the lack of permissions.

This comes with the tradeoff that users who _are_ logged in and _are_ in the correct team will experience a slight delay when they first load the app. It takes a noticeable amount of time for the app to recognize that they have access and enable the inputs. @kelshmo @Aryllen if you have time would you be able to try testing this out locally (or I can demo for you) and let me know if you think that's too much friction for the logged in user? The alternative is to keep inputs enabled on startup, and only disable them if they're not a member of the team. The disadvantage to that is the non-logged-in user would still have the options enabled.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,6 +26,7 @@ Imports:
     rlang,
     shiny,
     shinydashboard,
+    shinyjs,
     skimr,
     stats,
     synapser,
--- a/R/app-server.R
+++ b/R/app-server.R
@@ -22,15 +22,25 @@ app_server <- function(input, output, session) {
     membership <- check_team_membership(teams = c("3320424"), user = user)
     report_missing_membership(membership)
 
-    ## Create folder for upload (wrapped in try() to prevent app crashing when a
-    ## non-member of the team uses it, because otherwise the modal telling
-    ## people how to join isn't clickable)
-    created_folder <- try(
-      create_folder(
-        parent = "syn20506363",
-        name = user$userName
+    ## If user is a member of the team(s), create folder to save files and
+    ## enable inputs
+    if (inherits(membership, "check_pass")) {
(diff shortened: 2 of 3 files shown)


Aryllen:
I've not looked at the code itself, but on testing speed of allowing for use, this seems completely reasonable. At least locally, it only takes 2-3 seconds to activate the inputs. Outside of the most impatient users who are extremely familiar with the app, they will probably click once, at most, before it activates. An option could be to put a welcome message (e.g. "Logging in..." -> "Welcome, Dr. Quackenbush" or "Logging in..." -> "Sign into Synapse") somewhere (maybe the top bar) once it logs in. It would be a visual that would signal to users that it's ready for use.

kelshmo:
I don't think this is a noticeable wait time. As soon as I move to react - it is loaded. I can't imagine there would even be sufficient time to display an intermediate message. I have some specific comments I will address in the review.

PR #89, opened 2019-10-03
Allow checking IDs in only one direction

### LINKED ISSUE #85: Allow IDs in metadata that aren't in manifest
The app complains about specimen IDs that are in the metadata but not in the manifest. This type of mismatch will happen regularly (e.g. with additions to existing studies). Behavior should be as follows:

❌ mismatches between metadata files (e.g. between individual and biospecimen)
❌  mismatches where IDs are present in the manifest but not in the metadata
✅  mismatches where IDs are present in the metadata but not in the manifest

karawoo:
Fixes #85. This adds a `bidirectional` argument to `check_ids_match()` that defaults to `TRUE` (will check mismatches between the IDs in both directions) but can be set to `FALSE` in cases where we want to allow one set of IDs to be a superset of another (e.g. when someone is providing new data to an existing study and the manifest has fewer IDs than the comprehensive metadata).

The app now checks with `bidirectional = FALSE` when checking specimen IDs between the biospecimen file and manifest. If the manifest has specimen IDs that are missing from the biospecimen file, the check will fail; if the biospecimen file has IDs that are missing from the manifest, that's fine. One question I have is that in this case, the success message is the same as if the IDs were completely matching ("All specimenID values match between biospecimen and manifest") -- is that confusing, or fine?

### DIFF
--- a/R/app-server.R
+++ b/R/app-server.R
@@ -165,7 +165,13 @@ app_server <- function(input, output, session) {
       check_specimen_ids_match(biosp(), assay(), "biospecimen", "assay")
     })
     specimen_ids_biosp_manifest <- reactive({
-      check_specimen_ids_match(biosp(), manifest(), "biospecimen", "manifest")
+      check_specimen_ids_match(
+        biosp(),
+        manifest(),
+        "biospecimen",
+        "manifest",
+        bidirectional = FALSE
+      )
     })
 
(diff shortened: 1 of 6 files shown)


Aryllen:
Works as desired.

karawoo:
@Aryllen if this looks good to you will you approve it?

PR #92, opened 2019-10-07
Check if user is certified

### LINKED ISSUE #87: Check that user is certified
Realized this while looking into #86: a user could be signed in and part of the team, but not be a certified user, in which case the app will fail to create their folder or save their files. Should add a check (and yet another modal, or edit the existing one for team membership) to point out that they need to be a certified user. Need to do a REST call for this:

```
synRestGET('/user/{user_number}/certifiedUserPassingRecord')
```

karawoo:
Fixes #87. User must be certified to use the application since it saves info to Synapse. This adds another check and updates the modal message to include information about how to be come certified if needed. If the user is not certified _and_ not in the required teams, they will see something like this:

Screen Shot 2019-10-07 at 11 06 05 AM

If they have one or the other issue, they'll only see the message that pertains to them:

Screen Shot 2019-10-07 at 11 06 23 AM


karawoo:
@kelshmo does this seem ok to you? It will impact PEC as well, so just want to confirm.

kelshmo:
Yes, this is fine since contributors will need to be certified in order to upload data.

PR #97, opened 2019-10-09
Check for complete columns

### LINKED ISSUE #91: Figure out which columns need to be complete
As a possible stopgap before #72 we need to figure out which if any columns in the metadata and manifest must be complete (i.e. no missing values). I say this is a stopgap because the conditional logic in JSON Schema will cover this, but given the problem of using the Python validation library + synapser, this will take some time to implement.

Aryllen:
Fixes #91 (at least the start). This adds the `check_cols_complete()` function and testing. To be used for enforcing that certain columns are complete. Default is to return a `check_fail`, as opposed to a `check_warn`, if any of the `required_cols` are incomplete. 

Still need to figure out which columns to check for completeness and add this to the validator checks.


R/check-cols-complete.R
```diff
@@ -0,0 +1,50 @@
+#' Check for complete columns
+#'
+#' Check for complete columns in the data and warn (or fail) if present.
+#'
+#' @param data Data to check
+#' @param required_cols The required columns to check for completeness.
```
karawoo: Could we specify what format this should be in, to help the user? e.g. "A character vector of the required columns..."

R/check-cols-complete.R
```diff
@@ -0,0 +1,50 @@
...
+#' dat <- data.frame(specimenID = c("x", "y"), organ = c(NA, NA))
+#' check_cols_complete(dat, c("specimenID", "organ"))
+check_cols_complete <- function(data, required_cols,
+                                empty_values = c(NA, ""), strict = TRUE,
+                                success_msg = "Required columns are complete",
+                                fail_msg = "Some required columns are not complete") { #nolint
+  if (is.null(data)) {
+    return(NULL)
+  }
+  ## Check if all columns have data
```
karawoo:
```suggestion
  ## Check if columns in `required_cols` contain missing data
```

R/check-cols-complete.R
Aryllen: Thank you for the suggestion.

Aryllen:
Current change requests completed.

PR #102, opened 2019-10-11
Give test folder a timestamped name

### LINKED ISSUE #100: Intermittent isolated travis failures
I've been seeing this the last few weeks: sometimes one build in the matrix will fail in the `create_folder()` test; restarting the test always fixes it. I am guessing there's some issue with multiple builds being run at the same time and messing with the same Synapse folder; need to update this test to be more robust to that.

karawoo:
Hoping this will fix #100 but we'll have to wait and see to be sure. I'm not totally sure what has been causing these errors, but my hunch is that if multiple builds are running at roughly the same time they were interfering with the same folder. I can't force an error locally, however.

### DIFF
--- a/tests/testthat/test-create-folder.R
+++ b/tests/testthat/test-create-folder.R
@@ -6,9 +6,17 @@ attempt_login()
 test_that("create_folder() creates a folder", {
   skip_if_not(logged_in())
 
+  op <- options(digits.secs = 3)
+  ## Create folder name with timestamp including milliseconds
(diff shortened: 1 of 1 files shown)


karawoo:
It deletes them when it's done with this line:

https://github.com/Sage-Bionetworks/dccvalidator/blob/6d36b7f9a407c5ffe6f89393531f089df174531b/tests/testthat/test-create-folder.R#L22

PR #103, opened 2019-10-11
Custom file uploads by type

### LINKED ISSUE #58: Customize file upload requirements by data type
Drosophila studies only have biospecimen and assay metadata as it doesn't make sense to track information at the individual level. So for these studies, the individual metadata shouldn't be required, but all other checks (on the biospecimen/assay metadata and manifest) should remain the same.

Aryllen:
Fixes #58. 

- Moved species and assay options above `fileInput` boxes.
- Changed the species options to "human", "drosophila", and "mouse or other animal model."
- When "drosophila" is chosen, the `fileInput` for individual metadata is removed from the UI. If a user has already input individual metadata while a different species is chosen and then chooses drosophila, the `fileInput` and stored file data is reset. This makes any previous checks for the individual metadata disappear from the success/warn/fail boxes.
- Updates `check_cols_biospecimen()` to use the `general` template for humans. Also updated `app-server.R` to send in the species name to the check, instead of `general`.

Known bug:
If a user uploads individual metadata and then chooses drosophila from the species list, the drop-down menu in Data Summary does not remove the individual metadata file type (all of the other data summary information for individual metadata does clear as it should). If the user has uploaded any other data either before or after choosing drosophila, the individual metadata file type will disappear from the drop-down menu. The issue is in [this section](https://github.com/Aryllen/dccvalidator/blob/d16726b0b679e3726ca3bec04f06aec6985048da/R/app-server.R#L363-L404). 

I believe the problem lies in where `validate` is being called. Once drosophila is chosen, `indiv` becomes `NULL` and `validate` gives the error message instead of allowing `vals` to update. However, moving `validate` to some other place then makes the list creation for `vals` fail altogether since `indiv` (and the other datasets) doesn't yet exist. Still brainstorming how to fix the bug without getting too complicated, but this is ultimately a small bug that should not affect the average user experience.


Aryllen:
Fixed merge conflict.

Aryllen:
Oh my, I've no idea where those came from. Annoying. Will fix.

PR #106, opened 2019-10-14
Display the number of successes, warnings, and failures in title of boxes

### LINKED ISSUE #57: Include number of successes/warnings/failures in box header
Like

```
Successes (10)
--------------
Warnings (3)
------------
Failures (5)
------------
```

Aryllen:
Fixes #57. 

Not much to say that isn't said in #57. This is a small addition.


R/app-server.R
```diff
@@ -281,27 +281,46 @@ app_server <- function(input, output, session) {
...
         inherits(x, "check_fail")
       })
-      report_results(res()[failures], emoji_prefix = "x", verbose = TRUE)
+      output$failures <- renderUI({
+        report_results(res()[failures],
+                       emoji_prefix = "x",
+                       verbose = TRUE)
+      })
+      output$num_fail <- renderText({
+        paste("Failures (", as.character(sum(failures)), ")")
```
```suggestion
        paste0("Failures (", as.character(sum(failures)), ")")
```

R/app-server.R
```diff
@@ -281,27 +281,46 @@ app_server <- function(input, output, session) {
...
 
     ## Warnings box
-    output$warnings <- renderUI({
+    observe({
       warnings <- purrr::map_lgl(res(), function(x) {
         inherits(x, "check_warn")
       })
-      report_results(res()[warnings], emoji_prefix = "warning", verbose = TRUE)
+      output$warnings <- renderUI({
+        report_results(res()[warnings],
```
karawoo: Tidyverse style would put the `res()[warnings],` here on a new line

Aryllen: Thanks! I'll fix that and also install the dev version of styler. Perhaps it will help me catch these better.

Aryllen:
There _should_ be a way to put the title in the boxes upon startup, but I'm not sure how just yet. I'll look into it.

Aryllen:
Added a `reactiveValues` set that initializes the titles before checking for login requirements, and updates the string values to the have the current number of successes/warnings/failures in the titles when the boxes update.

R/app-server.R
```diff
@@ -283,27 +293,50 @@ app_server <- function(input, output, session) {
...
       successes <- purrr::map_lgl(res(), function(x) {
         inherits(x, "check_pass")
       })
-      report_results(res()[successes], emoji_prefix = "check")
+      output$successes <- renderUI({
+        report_results(res()[successes], emoji_prefix = "check")
+      })
+      reporting_titles$success <- paste0("Successes (",
+                                         as.character(sum(successes)),
+                                         ")")
```
karawoo: I've been using `glue()` in new code to try to make the construction of these messages a little more concise. The syntax here would be: `glue::glue("Successes ({sum(successes)})")` (`as.character()` isn't needed). Up to you whether you want to use this or not, just an idea, but either way will you please update the style of these lines? `"Successes (",` should go on a new line.

Aryllen: I think the `glue` option is better. I'll change it to that. It should make the line short enough to not need to drop down to a new line.

Issue #95, opened 2019-10-09
How-to guide for contributors


karawoo:
Create a document and/or screencast (I prefer the former as it's easier to keep up to date, but either or both is fine) documenting how data contributors can use the app.

kelshmo:
At what point do you plan to create a "stable" version of the dccvalidator so we can make relevant documentation? That is, if you plan on creating release versions of the app.

Agree that a document is preferred.

karawoo:
That is a good question. I think that once it's usable for PEC and we've addressed the important issues for the AMP-AD November data release, that would be a logical place for a v1.0.0. Though in practice, I don't expect releasing a major version to change our development practices much.

In terms of when to create this documentation, I think once #93 is merged #55 and #58 are complete we'll be in a state where all the big remaining user-facing features of the app have been completed. The other outstanding issues that would go into v1.0.0 are comparatively small. Given that #55 is blocked however, we may want to start on this documentation sooner. The main points (upload metadata files, see results, data summary tab, etc.) will be in place.

I'm thinking that writing this as a vignette makes sense: it'll live in the repo with everything else, we can render the rmd into a document to send collaborators, and we could create a pkgdown site for the whole project with this document as one page, which contributors could bookmark (advantageous if we make future changes and want to ensure they always have access to an up-to-date version of the doc).

karawoo:
Closed in #110

PR #110, opened 2019-10-15
Reorganize documentation with pkgdown + vignettes

karawoo:
@kelshmo this has a draft vignette for #95 and reorganizes the documentation to make the README more concise. It adds a [pkgdown](https://pkgdown.r-lib.org/) site which you can see locally with `pkgdown::build_site()`. @Aryllen and I discussed pros/cons of pkgdown -- the main downside I see is the potential to confuse data contributors, who really only need to view one vignette, with the rest of the content on the site. I tried it out just to see how it looks and think on it some more. Feel free to take the branch and make any updates you want, or go in a new direction entirely.

I wrote the vignette with language specific to AMP-AD, but we could remove that if we don't want to maintain separate sets of instructions. However the screenshots etc. will have the wrong URL for one or the other consortium, which might be confusing.


kelshmo:
@karawoo does `pkgdown` need to be specified in the `renv.lock` file to be able to run `pkgdown::build_site()`?

karawoo:
pkgdown needs to be installed, though not necessarily through the lockfile. By default renv only stores build- and deploy-time dependencies, not packages for interactive development (this is why lintr, styler, etc. are not included): https://rstudio.github.io/renv/articles/faq.html#how-should-i-handle-development-dependencies. I'm open to including these, but is has the potential to add a lot of extra stuff that not everyone who is cloning the project may need.

kelshmo:
Ok - I'm fine with not including it since it is only required for development. 

> the main downside I see is the potential to confuse data contributors

The availability of this documentation is important for many different end-users. What do you think about, for our data contributors, linking directly from the app to to the articles page - e.g. "Using the dccvalidator app in AMP-AD"?

karawoo:
Agreed we'd want to link them directly to the page of interest. They still might see other stuff that's irrelevant to them in the header bar. But I think that is probably ok, and preferable compared to emailing them a document that may become out of date and that they might lose.

kelshmo:
I have a draft vignette for PEC like your [AMP-AD draft](vignettes/using-the-dccvalidator-app-amp-ad.Rmd). What is the best way to contribute this doc? Open a new PR with only the additional .Rmd document? I would love to get some feedback even if the PR remains a draft for the moment. I can host it on Synapse for the face to face tomorrow so there is not pressure to get this online.

karawoo:
Great! Do you want to open a new PR against this branch?

karawoo:
@kelshmo I added the images and re-rendered everything, will you take a look and let me know if you think we're ready to merge?

karawoo:
ugghhh I thought I solved the images issue but apparently not

karawoo:
> Images look great! I see 4 rendering on the AMP-AD vignettes with nice drop shadow. Ready to merge - I will add screenshots for PEC in another PR.

The problem was they were rendering in the pkgdown site, but not when Travis tried to build the vignettes (hence the failing build). I've pushed a change that I hope will fix it 🤞

PR #117, opened 2019-10-16
Open files with `stringsAsFactors = FALSE`

### LINKED ISSUE #115: Load data with stringsAsFactors = FALSE
Not currently causing any issues, but it would be nice to have the data summary legend show "character" instead of "factor", as the former is more understandable.

Screen Shot 2019-10-15 at 4 50 53 PM

Aryllen:
Fixes #115.

Not much to say other than all the files now load in with `character` instead of `factor` for strings. This displays correctly in the Data Summary.


karawoo:
@Aryllen I notice a lot of extra commits in these PRs. For future branches, could you please ensure you have an up-to-date version of master and create the branch off of that? I can show you how if needed.

PR #119, opened 2019-10-17
Check complete cols

### LINKED ISSUE #91: Figure out which columns need to be complete

Aryllen:
Fixes #91.

Added in the columns that need to be complete based on the list in #91. 

Changed logic for `check_cols_complete()`. The logic had assumed all `required_cols` were present, which may not be the case. Removed this assumption. The function now checks for any missing `required_cols` and if any of the existing `required_cols` has missing data. If there are no missing `required_cols` in the data and all existing `required_cols` have complete data, the function returns a `check_pass`. If either of those conditions is false, it returns a `check_fail` with the names of the columns missing altogether and/or the columns missing data.

Added a couple test cases for this function. While it made the test overly long, I put them in the section "check_cols_complete returns check_fail if any required column lacks data" simply because that's still the case it's checking for.


R/check-cols-complete.R
```diff
@@ -23,20 +23,25 @@ check_cols_complete <- function(data, required_cols,
   if (is.null(data)) {
     return(NULL)
   }
-  ## Check if columns in `required_cols` contain missing data
+  req_cols <- required_cols[required_cols %in% names(data)]
```
karawoo: It's a small thing but could we give this a more informative name? I'm already mixing up `required_cols` and `req_cols` as I'm reading the rest of the code. Maybe `req_cols_present` or something?

R/check-cols-complete.R
```diff
@@ -23,20 +23,25 @@ check_cols_complete <- function(data, required_cols,
...
     paste(required_cols, collapse = ", "),
     " should be complete."
   )
 
-  ## Return success if all required columns have complete data.
+  ## Return success if all required columns have complete data,
+  ## and there are no missing required columns.
   ## Otherwise return warn or fail depending on `strict` argument
-  if (!any(results)) {
+  if (all(!results) & length(missing_cols) == 0) {
```
karawoo: Is there a need to change `!any(results)` to `all(!results)` here?

R/check-cols-complete.R
Aryllen: For sure! I have a tendency towards overly long variable names. Restraining myself here made this section a bit confusing. Haha.

R/check-cols-complete.R
Aryllen: Nope. I changed it back to `!any(results)`, which should be a shorter runtime. Thank you for catching that!

Long, uninteresting explanation: I was getting the wrong results and was thinking it was a logic issue, not a forgetting-to-reload-package-before-testing issue. Tried out multiple different things, including changing this bit to an equivalent condition in case there was something odd with the ordering. Forgot to switch it back after I fixed the problem (i.e. reloaded the package with the newest version of the function).

Aryllen:
I just realized the documentation for this function was a little incorrect. We give a fail (or warn) if the columns are incomplete, not the other way around. I also put a line explaining how missing `required_cols` are treated.

R/check-cols-complete.R
karawoo: makes sense, thanks!

PR #121, opened 2019-10-17
Add annotations, including study name, to metadata & manifest files

### LINKED ISSUE #109: Add study name annotations for metadata
To track which study the metadata files belong to, we would need to store the study name. This would help with creating a monitoring app that checks validation progress for each study. As it is now, there are no ties between files and studies other than the user Id of the person validating. Since there could be multiple studies that a single person is validating data for, this would make it difficult to simply infer the study.

This would require adding in the `get-study` module to the validator and getting the study name from the user before validating (i.e. `fileInput` features would be deactivated until a study name was given).

Aryllen:
Fixes #109.

Added the `get-study` module to Validator tab. Requires the user to give a study name before the other UI elements become enabled.

When files are uploaded for validation, they are saved with annotations depending the file.
- `study` name: all files
- `metadataType`: all metadata files
- `assay` name: assay files


Aryllen:
Realized there was a problem with logic for enabling the UI input elements (needed to disable the UI input elements under certain conditions). Latest two commits fix the issue.

karawoo:
Similar to #122 I'm a little worried about the user experience of having so much of the page disabled without explaining why. I think maybe there should either be some text that explains what they need to do, or all the inputs that are disabled should be hidden until they've entered the study. What do you think?

Aryllen:
The challenge is that the section for getting the study does not go away (or become inactive) once a study name is given. Having it deactivate would be more challenging due to how it reactively updates (for example, if someone were to type a name in the textbox, then the moment the reactive input updated, it would disable the textbox regardless of whether they were finished typing).

We could have the fileInputs hidden until a study name is given, but this also brings challenges due to the reactive updating. The user could have the updates pop up before they've selected/entered the correct group name. If they selected a dropdown name by accident/purpose and then selected `No` for the study existing, then the fileInputs would disappear again until they entered a name. This behavior might be confusing.

One thought I had was to give the study name module a button. Once they've given the study name, they click the button and the name doesn't update until they click it. This would reduce the reactive issues. We could also see about having it become hidden once they've clicked it. That would remove the ability to change the study name after clicking. The fileInputs could be hidden during this process and become visible when the study name is accepted, as well. In the future, we could even add a "reset" button that resets everything back to the original settings (and brings back the study module), if there appears to be friction with not being able to get back to the original state.

Another thing I thought of was potentially having the first tab be only for getting the study name, and the other two tabs being disabled unless the study name is given. The study name wouldn't have to be entered for the other two sections, which reduces redundancy. That might be a bit more challenging to implement, however.

Aryllen:
I'm expecting a lint error on the lastest push. 

Per the conversation this morning, I changed this addition to have a `validate_btn` that uploads the files with the annotations to Synapse and populates the success/warn/fail boxes. It will give an error message if the user does not submit the name or at least one file. All of this appears to be functioning correctly.

The bad part is that moving all this to an `observeEvent` for `validate_btn` increases the cyclomatic complexity of the full `app-server()` (hence, the expected lint error). This could potentially be mitigated with changing the validation section into modules that work together.

Aryllen:
The new change added a bug. Will update when I fix it.

Aryllen:
Fixed bug. Didn't update the variable name correctly when I moved the uploading code to a different section.

karawoo:
@Aryllen would you mind fixing the merge conflict before I re-review?

I think we can ditch the cyclomatic complexity linter for the time being, and potentially re-enable once the app is more modularized. Adding `cyclocomp_linter = NULL` to the first line of the .lintr file should take care of this.

Aryllen:
Disabled linting for cyclomatic complexity and took care of conflict.

R/app-server.R
```diff
@@ -378,45 +366,109 @@ app_server <- function(input, output, session) {
...
+        ## Require at least one file input
+        validate(
+          need(
+            any(
+              !is.null(indiv()),
+              !is.null(biosp()),
+              !is.null(assay()),
+              !is.null(manifest())
+            ),
+            message = "Please upload some data to view a summary"
```
```suggestion
            message = "Please upload some data to validate"
```

Aryllen: This message is in the data summary section. I don't believe it's used elsewhere. This is what that section would look like if we changed the message. It makes some sense, but not as much sense as what it is now.

image

Aryllen:
I made a simple file from the individual metadata template with some nonsense in the rows. This was used for all three metadata files since it doesn't necessarily matter what is in those. The manifest was the one I'm using normally, but with the lines copy/pasted enough times. I ran the test multiple times, three times from cold starts (first time after opening the app) and more from warm starts (after having uploaded once). The timing isn't exactly precise; I rounded to the nearest second. Overall, it doesn't appear to be much difference between large files and small files. 

image

Some things I did notice:
- If a file has been uploaded in a session, it's not re-uploading the files to Synapse if it's the exact same file, only if the file has been changed/different. This may be what is causing the difference between cold starts and warm starts.
- The titles of the success/warn/fail boxes now take longer to appear. I believe this is due to the ordering of the tabs now. The items in Documentation would load first, making it take longer to load the items in Validator. Again, I will check into this to see if there is a way to make them appear quicker.

R/app-server.R
karawoo: The data summary message is elsewhere, this line causes the message to appear if you try to click validate without submitting any data.

Screen Shot 2019-10-23 at 9 05 31 AM

Aryllen: Oh! Haha, nice. Yeah, that definitely makes sense to change. Thanks for catching that!

Aryllen:
I'm not sure how to get the success/warn/fail boxes to load the titles faster. They _should_ load first, but I think the `observeEvent` for the cookie takes over too soon, pushing the box title's `renderText` until after everything else is loading. At least that is my hypothesis. If we change to Oauth and modules, we might have more wiggle room, but I cannot currently figure out how to force the titles to appear before the cookie logic starts.

PR #122, opened 2019-10-17
Require study name in documentation

Aryllen:
This fixes a problem I discovered with the study name. The dropdown box populated with the study name list and the first value (ACT) was selected upon loading. If a user uploaded documentation without touching the study name module, then the documentation would be annotated with the wrong study name. Additionally, if they selected `No` from the `study_exists` radiobuttons, but did not enter a name in the textbox, the value would still be ACT.

This update forces the user to give a study name to be able to use the `fileInput` UI elements.
- Changed the first element in the study name list to be "".
- Made the study name reset to "" if `No` is selected from the `study_exists` radiobuttons.
- Made it so the `fileInput` UI elements only become enabled after the study name is given.
- `fileInput` UI elements disable if the study name is the default of "".


Aryllen:
I thought of that after doing this. Haha. For the documentation section, it would be easier to simply keep the submit button disabled until the study name is given. This won't quite work for the validation section, however, since there isn't a submit button for those files. Will change this after we discuss how to handle the validator section since it might affect the documentation portion, as well.

Aryllen:
Per our conversation this morning, I changed this update to give an error message if the `submit` button is pressed without giving the study name.

PR #123, opened 2019-10-17
Set app options from a configuration file

### LINKED ISSUE #69: Make app configurable for PsychENCODE
Right now the process to repurpose the app is:
* Fork the git repo
* Modify the contents of the server and UI functions to
  * Save uploaded files to a different location
  * Customize the checks that are run
  * Customize inputs if needed
* Deploy in a new location on the shiny server

This is doable, but it would be better to make the overall app more configurable. A first step is likely splitting the components into shiny modules (which would be a best practice regardless of whether the app needs to be reused for other projects).

karawoo:
Fixes #69 by introducing a configuration file to customize Shiny app options. The configuration file currently contains the following options:

`storage_location`: synID of location where files will be stored
`study_table`: synID of table listing existing studies in the project
`annotations_table`: synID of table of allowable annotation values
`annotations_link`: link to annotations dictionary (shown in instructions)
`templates_link`: link to where templates are hosted (shown in instructions)
`teams`: list of teams user must be a member of one of to use the application
`templates`: named list of template synIDs. Has sub-options `manifest_template`, `individual_templates`, `biospecimen_templates`, `assay_templates`.
  manifest_template: "syn20820080"
`species_list`: list of species to show as user input options

I have created 3 configurations: one "default" that stores data to a test folder, one for AMP-AD (which inherits most of its other options from the default), and one for the PsychENCODE customizations I am aware of.

In addition, I have changed the behavior of `check_cols_biospecimen()` etc. to look for their templates by synID rather than name, so PEC does not need to edit these package functions.

With these changes in place, the workflow for creating a custom version of the app would be:
1. Fork this repo
2. Create a Synapse folder with the appropriate permissions to store uploaded files
3. Add your customization to `config.yml` (overwriting existing customizations if desired, but a "default" customization is required).
4. Set the active configuration in app.R with `Sys.setenv(R_CONFIG_ACTIVE = "your_config_here")`

The configuration file does not alter which checks are run, or with which arguments. Though the config file can contain R code, this code is long and complex enough that I do not believe it would be advisable to put it in the config. Perhaps instead there should be separate modules for AMP-AD's checks and PEC's checks, and the config.yml file could indicate which module to call.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -13,3 +13,4 @@
 ^\.lintr$
 ^\.pre-commit-config\.yaml$
 ^\.Rprofile$
+^config\.yml$
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -19,6 +19,7 @@ Depends:
     R (>= 3.4),
     shinyBS
 Imports:
+    config,
     DT,
     emo,
     ggplot2,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dccvalidator
 Title: Validate CNS Data and Metadata
-Version: 0.0.0.9005
+Version: 0.0.0.9006
 Authors@R: 
     c(person(given = "Kara",
              family = "Woo",
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dccvalidator
 Title: Validate CNS Data and Metadata
-Version: 0.0.0.9006
+Version: 0.0.0.9007
 Authors@R: 
     c(person(given = "Kara",
              family = "Woo",
--- a/R/app-server.R
+++ b/R/app-server.R
@@ -43,7 +43,7 @@ app_server <- function(input, output, session) {
       inherits(certified, "check_pass")) {
       created_folder <- try(
         create_folder(
-          parent = config::get("storage_location"),
+          parent = config::get("parent"),
           name = user$userName
         )
       )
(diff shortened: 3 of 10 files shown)


karawoo:
> Would it be useful to add the list of columns that are required to be complete to this? It's possible that the columns would be the same across consortium, but if there was a chance it wasn't, then this would also help with configuring for a specific group.

This will definitely need to be customizable one way or another. I was thinking if we turn all the checks into a couple of consortium-specific modules, this information would go in there. But I'm open to having it in the config instead.

Thanks for reviewing, I will hold off on merging until @kelshmo has a chance to look.

karawoo:
Note from @philerooski to rename `storage_location` to something like parent, since storage location is a distinct synapse concept

karawoo:
Note from @kdaily: make setting the active config conditional on it not already being set.

karawoo:
Note to self: add required columns to config.yml (right now this is the only customization PEC is making to the checks that are run, so we can add that as a config for now and worry about pulling it out into separate function(s) later)

kelshmo:
With `Sys.setenv(R_CONFIG_ACTIVE = "pec")` I'm seeing an error with `annotation_values_indiv` in `app_server.R`. The file is still uploading to Synapse successfully, but the app is failing. The error is `Warning: Error in if: argument is of length zero `.

Additionally, proteomics is still populating from somewhere. Could it have something to do with how default arguments are handled? 
Screen Shot 2019-10-24 at 10 48 11 AM

I'm going to look into this further.

PR #126, opened 2019-10-18
Explain how to skip pre-commit hooks if needed

karawoo:


karawoo:
No worries, I tagged @kelshmo because she and I had already had a conversation that prompted this PR, but I always welcome you adding a review as well even if you're not explicitly tagged--the more eyes on it, the better!

PR #141, opened 2019-10-25
Fix NA and bidirectional issues in check_ids_match.

### LINKED ISSUE #135: Fix NA and bidirectional bug in matching IDs
`check_ids_match` should not be treating `NA` as an IndividualID.  This is an issue for specimenIDs, as well.

It also appears that bidirectional may have a bug as it shouldn't be checking which IndividualIDs are missing from the manifest. In the tested file, none were missing, but there appears to be an error with how the file is written or read in. This file error brought the bidirectional issue to light.

image

Aryllen:
Fixes #135.

First issue was that `NA` was considered an individualID or biospecimenID. The manifest will have some columns that are `NA` due to the metadata files. These should be considered okay. Additionally, if one ID column is simply incomplete in a biospecimen or individual file (i.e. has some `NA`), then it will be flagged by the incomplete required column check anyway.

Second issue was that the `bidirectional = FALSE` flag checks the second dataset against the first dataset. The original order was checking the manifest IDs to see if any were missing from the biospecimen/individual IDs. It now checks to see if any of the biospecimen/individual IDs are missing from the manifest, which is as intended.

### DIFF
--- a/R/app-server.R
+++ b/R/app-server.R
@@ -220,8 +220,8 @@ app_server <- function(input, output, session) {
     })
     individual_ids_indiv_manifest <- reactive({
       check_indiv_ids_match(
-        manifest(),
         indiv(),
+        manifest(),
         "individual",
         "manifest",
         bidirectional = FALSE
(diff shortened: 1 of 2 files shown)


Aryllen:
I had this big long explanation for how I was thinking about the problem and then realized I had it backwards. The `manifest` can have fewer `individualID`s if uploading additional data for an existing study, but the `individual` metadata needs to have all listed. The commit changing the order is reverted.

PR #142, opened 2019-10-25
Fix check_cols_complete to not fail for missing required columns

### LINKED ISSUE #136: Complete required columns check is wrong for drosophila metadata
Currently requiring individualID and biospecimenID in the biospecimen metadata. However, there is no individualID for drosophila metadata.

Some options:
1) Make a more complicated config file that tailors required columns by template.
2) Make some other check in the R code that tailors required columns in config file by template.
3) Change `check_cols_complete` to only check required columns that exist in the data.

Aryllen:
Fixes #136.

Make `check_cols_complete` ignore required columns that do not exist in the dataset. Update documentation for function. Remove tests specific to checking that it was failing for missing required columns.

This problem comes up in Biospecimen since we require `individualID` and `specimenID` to be complete columns. If the species is drosophila, there will be no `individualID`. The current version would then fail the check due to the column being missing. This version would pass (assuming `specimenID` was complete).


Aryllen:
Updated the success message and added an extra test for `check_pass` if required columns are missing. I left the other tests removed since they were inappropriate for testing the new behavior.

karawoo:
Could you run `devtools::document()` again and push the updated documentation?

Aryllen:
Whoops, done. Thanks!

PR #171, opened 2019-10-29
Give error if study name does not fit filename requirements

### LINKED ISSUE #152: Check for invalid Synapse study names
A user during the Bug Bash session said, "Putting an invalid Synapse name in Study name shows Synapse 400 error."

Since the study name is appended to the front of the documentation filenames, it has to be something that is valid within a Synapse filename.

Aryllen:
Fixes #152.

Since study names are prepended to the documentation filenames, the study name cannot have certain special characters. This update creates a temp string with the allowed special characters removed. Then checks if any punctuation characters still exist in the string.


R/mod-upload-documentation.R
```diff
@@ -151,6 +151,15 @@ upload_documents_server <- function(input, output, session,
         if (study_name() == "") {
           stop("Please enter study name.")
         }
+        # Check if study name has inappropriate characters
+        temp_string <- stringr::str_replace_all(
```
karawoo: Could you add stringr under Imports in the DESCRIPTION file?

Aryllen: Done.

Side note: the commit for this change had the pre-commit hooks turned off since 'use-tidy-description' failed every time. It doesn't fail if I run `usethis::use_tidy_description()' from RStudio within the project. The only thing I can think of is the fact that it prints output saying that it's changing the active project to dccvalidator (it is the active project so that's odd regardless). The output might be being flagged as an error.

Aryllen:
We could enforce it in the validator. It is less important at the moment since, as you said, we are not changing the filenames there. It definitely doesn't hurt, though. I can add that check there, but I have a question about preference on how to implement that since it's the same code. 

Would you prefer that I simply add the few lines to validator or add a separate `study_name_check` function that is called both times?

karawoo:
Cool, yeah I think it would be better to be consistent. 

> Would you prefer that I simply add the few lines to validator or add a separate study_name_check function that is called both times?

I think a separate function would be good (and it's not shiny-specific, so we can add tests for it!)

Aryllen:
> I think a separate function would be good (and it's not shiny-specific, so we can add tests for it!)

Definitely good to be able to add tests! I'll update this to a function with testing and comment when the latest update is ready for review, again.

Aryllen:
Added everything discussed. I made `is_study_name_valid()` a simple true/false function. The areas that use it have an if statement that sends a stop error with message if the name is invalid.

tests/testthat/test-is-study-name-valid.R
```diff
@@ -0,0 +1,41 @@
...
+  expect_true(is_study_name_valid(name2))
+  expect_true(is_study_name_valid(name3))
+  expect_true(is_study_name_valid(name4))
+  expect_true(is_study_name_valid(name5))
+  expect_true(is_study_name_valid(name6))
+  expect_true(is_study_name_valid(name7))
+  expect_true(is_study_name_valid(name8))
+})
+
+test_that("is_study_name_valid returns FALSE when name is invalid", {
```
karawoo: Can we add one test for the study name `""`?

Aryllen: Haha, whoops. Thank you for catching that! Added.

PR #174, opened 2019-10-30
Add scRNAseq, WGS, WES, and NanoString templates

### LINKED ISSUE #90: Add support for new metadata templates
WGS, WES, nanostring, single cell RNA seq, and possibly others are expected for November.

Need to add support for these once they have been shared in https://www.synapse.org/#!Synapse:syn18512044

karawoo:
~Part of #90 (but does not complete it as we are still waiting for nanostring)~
Closes #90

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -40,6 +40,9 @@ amp-ad:
   templates:
     assay_templates:
       proteomics: "syn12973255"
+      single cell RNAseq: "syn21018360"
+      wholeGenomeSeq: "syn21018358"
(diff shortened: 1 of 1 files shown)


karawoo:
Yes, I think we agreed these assays would use the same templates. Is that aligned with what you do in PEC?

karawoo:
Though maybe saving a renamed copy of the file is more logical for the user when they go to the folder and view the templates.

kelshmo:
I don't think we have curated WES data yet and I agree a renamed template is preferred.

PR #184, opened 2019-11-06
Deploy pkgdown site with Travis

### LINKED ISSUE #177: Build pkgdown site automatically with Travis
Might need to put it in gh-pages instead: https://pkgdown.r-lib.org/reference/deploy_site_github.html

karawoo:
Closes #177. I more or less followed the setup instructions [here](https://pkgdown.r-lib.org/reference/deploy_site_github.html), hopefully this works? After this is merged we'll need to update the github settings to serve the site from the gh-pages branch instead of from the docs/ folder on master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,11 @@ jobs:
         - |
           R CMD INSTALL $PKG_TARBALL
           Rscript -e 'lintr::lint_package()'
+    - stage: deploy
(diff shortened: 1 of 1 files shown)


karawoo:
Currently it's the old site that's still being served. We will know this works if it pushes to the `gh-pages` branch after it's merged. At that point we can remove everything in the docs/ folder on the master branch, and change the GitHub settings to serve from the gh-pages branch instead.

PR #191, opened 2019-11-08
Check file names valid

Aryllen:
I don't believe we opened an issue for this, but it was mentioned that we should check that the filenames are valid for Synapse.

Changes:
- check-study-name.R file becomes check-name.R
- `is_study_name_valid()` becomes `is_name_valid()`
- `is_name_valid()` becomes more generic
- documentation updated
- tests updated to reflect new function and file names
- testing for study name validation changed from old function name to new function name
- added a test before each file upload that checks if the filename is valid

Note that the check for filename validity is done on a one-by-one basis. This means that the first file could upload fine if the name passed, but the second file would fail to upload (along with any others after it since it would stop). I considered checking all names at once, but we don't always have all files since we allow them to check any number of the files. By checking one-by-one, this allows for skipping the check if the variable is `NULL`.


R/app-server.R
```diff
@@ -385,6 +385,9 @@ app_server <- function(input, output, session) {
 
         ## Upload only the files that have been given
         if (!is.null(indiv())) {
+          if (!is_name_valid(files$indiv$name)) {
+            stop("Please check that file names only contains: letters, numbers, spaces, underscores, hyphens, periods, plus signs, and parentheses.") # nolint
+          }
```
Aryllen: This is a lot of repetition, with one per file to be uploaded. Doing this for the documentation filenames is going to be a hassle, as well, due to the unknown number of files (a user can upload 1+ assay documentation files). What do you think about adding this check to `save_to_synapse()`?

Aryllen:
I realized that I missed the filename checks for documentation, and was thinking on the repetition. If we add the check to `save_to_synapse()`, it would cover all files that are uploaded, even when we add future functionality for multiple assay files.

R/app-server.R
karawoo: Yeah that seems good!

Aryllen:
I updated this branch with the following changes:
- `is_valid_name()` checks for `NULL` now.
- `is_valid_name()` checks were nested in an `if-else` so that the test would exit earlier if the name was `NULL` or `""`. Before, the test would continue to check the string, but that was unnecessary and could have caused issues. Instead of using multiple `return` statements, I chose to use `if-else` for the logic.
- added test for the name being `NULL` in `test-is-name-valid.R`.
- moved the filename check from the server function to `save_to_synapse()`.

R/utils.R
```diff
@@ -52,6 +52,9 @@ save_to_synapse <- function(input_file,
                             parent,
                             name = NULL,
                             annotations = NULL) {
+  if (!is_name_valid(name)) {
+    stop("Please check that file names only contains: letters, numbers, spaces, underscores, hyphens, periods, plus signs, and parentheses.") # nolint
+  }
   file_to_upload <- synapser::File(
     input_file$datapath,
     parent = parent,
```
karawoo: I think we should change this function signature a little bit. It doesn't seem right to have a default value that we then say is invalid (even though in practice we usually override the default value). And the reason `name` defaults to `NULL` in `save_to_synapse()` was because I thought that was the default value of `File()`, but it turns out when you explicitly pass `NULL` to `File()` you change the behavior: https://sagebionetworks.jira.com/browse/SYNR-1437.

What do you think about removing `name` as an argument altogether, since we seem to always want to get it from the input file anyway?

```r
save_to_synapse <- function(input_file, parent, annotations = NULL) {
  file_to_upload <- synapser::File(
    input_file$datapath,
    parent = parent,
    name = input_file$name,
    annotations = annotations
  )
  if (!is_name_valid(file_to_upload$properties$name)) {
    stop("Please check that file names only contains: letters, numbers, spaces, underscores, hyphens, periods, plus signs, and parentheses.") # nolint
  }
  synapser::synStore(file_to_upload)
}
```

R/check-name.R
```diff
@@ -0,0 +1,25 @@
+#' Checks that the name of the study or file is valid.
+#' Valid strings contain: letters, numbers, spaces, underscores,
+#' hyphens, periods, plus signs, and parentheses.
+#'
```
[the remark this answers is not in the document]
Aryllen: What does this do?

R/utils.R
Aryllen: That makes sense. My only question is that that changing the `save_to_synapse()` signature is a different issue and involves more changes that simply checking that the file name is valid. What do you think about me fixing the call to `is_name_valid()` to use the name from the `input_file` and then opening a separate issue to fix `save_to_synapse()`?

R/check-name.R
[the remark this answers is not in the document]
karawoo: Marking the function as internal removes the function from the package index. It's not too important, but it's something that pkgdown was complaining about when I built the site locally, because we have functions that are documented but not listed on the Reference page (because they're not exported).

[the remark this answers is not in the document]
Aryllen: Ah, okay. I wanted to ask before I committed the suggestion in case I forgot about it later. Haha. Thanks for the explanation!

R/utils.R
karawoo: Sounds good to me, thanks!

Aryllen: Commit 8cca7e9 changes the call to `is_name_valid()` in `save_to_synapse()` to use `input_file$name` in preparation for changing the signature for `save_to_synapse()`. 

I left the test for a `NULL` name in both the `is_name_valid()` function and test-check-name.R. Did you want me to remove these checks? With how we are using this function, there should not be a case where `NULL` is sent to `is_name_valid()`, but there could be an edge case that I am missing.

R/utils.R
```diff
@@ -52,6 +52,9 @@ save_to_synapse <- function(input_file,
                             parent,
                             name = NULL,
                             annotations = NULL) {
+  if (!is_name_valid(input_file$name)) {
+    stop("Please check that file names only contains: letters, numbers, spaces, underscores, hyphens, periods, plus signs, and parentheses.") # nolint
```
karawoo:
```suggestion
    stop("Please check that file names only contain: letters, numbers, spaces, underscores, hyphens, periods, plus signs, and parentheses.") # nolint
```

PR #192, opened 2019-11-08
Update `check_empty_cols()` to ignore required columns.

### LINKED ISSUE #162: Should something be listed as both a warning and failure?
image

These are two separate tests, one that just gives a warning for empty columns and the other gives a fail because `parent` is required to be complete.

Aryllen:
Fixes #162.

Updates `check_empty_cols()` to take a parameter `required_cols` similar to `check_cols_complete()`. `check_empty_cols()` then ignores/doesn't check the columns in `required_cols` for emptiness.

Changes:
- Add optional `required_cols` parameter to `check_empty_cols()`, and have it ignore columns in this list.
- Update documentation.
- Update tests for `check_empty_cols()` to make sure it's removing `required_cols` from the check.
- Update the metadata checks to pass in `required_cols`.

### DIFF
--- a/R/check-cols-empty.R
+++ b/R/check-cols-empty.R
@@ -5,7 +5,7 @@
 #' not tested for emptiness. This is due to the existing function
 #' `check_cols_complete()`, which ensures that the required columns are
 #' complete. By ignoring the required columns in `check_cols_empty()`,
-#' there are no conflicting results for the same column in the event
+#' there are no duplicated results for the same column in the event
 #' that a required column was also empty.
 #'
 #' @param data Data to check
--- a/R/check-cols-empty.R
+++ b/R/check-cols-empty.R
@@ -8,20 +8,27 @@
 #' @param strict If `FALSE`, return a `"check_warn"` object; if `TRUE`, return a
 #'   `"check_fail"` object
 #' @inheritParams check_values
+#' @inheritParams check_cols_complete
 #' @return A condition object indicating whether the data contains columns that
 #'   are empty.
 #' @export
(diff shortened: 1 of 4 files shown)


Aryllen:
I've updated the documentation for the function. I wasn't sure how much information was too much, but I tried to be clear and concise. Feel free to edit if there's a better way to explain it.

R/check-cols-empty.R
```diff
@@ -1,27 +1,40 @@
 #' Check for empty columns
 #'
 #' Check for empty columns in the data and warn (or fail) if present.
+#' The function takes in a list of required column names that are
+#' not tested for emptiness. This is due to the existing function
+#' `check_cols_complete()`, which ensures that the required columns are
+#' complete. By ignoring the required columns in `check_cols_empty()`,
+#' there are no conflicting results for the same column in the event
```
karawoo:
```suggestion
#' there are no duplicated results for the same column in the event
```

Aryllen: Thanks!

Aryllen:
I just noticed that the Travis build failed on this PR. Based on the details it gives, I'm not sure what went wrong.

karawoo:
Looks like travis encountered an issue installing R, nothing wrong with our code. I restarted that build job to be sure. However could you re-run `devtools::document()` before we merge so the wording change to the documentation is updated?

Aryllen:
Documentation pushed. I totally forgot that the change was a documentation one, which means I needed to pull the changes and run the tool.

PR #193, opened 2019-11-11
Add note about expected files for documentation

### LINKED ISSUE #159: Add notes about file requirements
Copy/pasted from Bug Bash document due to being multiple things.

> Add a note about study and assay description file format (mette)
> In addition to the above a limit to the file size 
> And, file name: Names may only contain: letters, numbers, spaces, underscores, hypens, periods, plus signs, and parentheses

Not completely sure if some of these things would/should only refer to documentation files or for all files.

Aryllen:
Fixes #159. 

Updated:
- The documetation `fileInput`s expect .txt, .docx, .md, .pdf, and .tex. Note that this doesn't force the document upload to be one of these types; it simply makes the file explorer show these file types versus all file types.
- Added a note in the label for each documentation `fileInput` that gives the expected file types.

### DIFF
--- a/R/mod-upload-documentation.R
+++ b/R/mod-upload-documentation.R
@@ -25,14 +25,16 @@ upload_documents_ui <- function(id, study_link_human, study_link_animal) {
         shinyjs::disabled(
           fileInput(
             ns("study_doc"),
-            "Upload the study description file"
+            "Upload study description file (.txt, .docx, .md, .pdf, .tex)",
+            accept = c(".txt", ".docx", ".md", ".pdf", ".tex")
           )
         ),
         shinyjs::disabled(
           fileInput(
             ns("assay_doc"),
-            "Upload the assay description files",
-            multiple = TRUE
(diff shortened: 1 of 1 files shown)


R/mod-upload-documentation.R
```diff
@@ -25,14 +25,16 @@ upload_documents_ui <- function(id, study_link_human, study_link_animal) {
         shinyjs::disabled(
           fileInput(
             ns("study_doc"),
-            "Upload the study description file"
+            "Upload study description file (.txt, .docx, .md, .pdf, .tex)",
+            accept = c(".txt", ".docx", ".md", ".pdf", ".tex")
```
karawoo: I think we need the MIME types here, not just the file extensions.

Aryllen: Oh! That's interesting. It seems to work with the extensions. I changed to the MIME types for these. Thanks for catching that!

karawoo: Oh huh, the documentation says it needs the MIME type but maybe that's not true anymore. I tested your update and it does work as well

PR #194, opened 2019-11-12
Link vignette to app

kelshmo:
While the markdown is rendering in the app there are still some kinks to work out: 
- ~There is an output `.md` file in the root directory. I'm looking into specifying the output file path such that the markdowns are stored nested in `/vignettes`~
- ~Images are not rendering~
- ~Level 4 headers are not correct, aesthetically. H4 is smaller than the text itself.~


kelshmo:
@karawoo Knitting a markdown and storing the output in a nested directory is of course [more complicated than it should be](https://gist.github.com/jennybc/362f52446fe1ebc4c49f). Do you have a preference for dealing with the directory structure?

karawoo:
I'm chuckling because I taught this exact topic at R-Ladies this weekend. The `here` package is the best solution to creating paths to things that I know of. I don't have very strong feelings about where the generated .md files are stored (presumably they won't be checked in to git?)

kelshmo:
I figured you would have the expert knowledge :) I tested out `here` and was not able to configure it in the project (though, it doesn't sound like we need to worry about a nested folder).

Just adding `here()` to `app-server.R` returned the following error:
```
  output$markdown <- renderUI({
    HTML(markdown::markdownToHTML(
      knitr::knit(here::here(input = config::get("path_to_markdown"),
                  quiet = TRUE)),
      stylesheet = ("inst/app/www/custom.css")))
  })
#Warning in file(con, "r") :
#  cannot open file '/Users/kmontgomery/Documents/dccvalidator/vignettes/using-the-dccvalidator-app-#amp-ad.Rmd/TRUE': Not a directory
#Warning: Error in file: cannot open the connection
```

karawoo:
You have a misplaced closing paren I think, does it work if you do this?

```r
  output$markdown <- renderUI({
    HTML(markdown::markdownToHTML(
      knitr::knit(here::here(input = config::get("path_to_markdown")),
                  quiet = TRUE),
      stylesheet = ("inst/app/www/custom.css")))
  })
```

kelshmo:
Now that the markdown is in the correct directory, the images are rendering. (but HUGE!)

kelshmo:
> One is that by rendering the Rmd files in the server, we increase the time it takes for the app to load (the files will get rendered each time someone connects). Since these documents are fairly static, we might be able to speed things up by rendering the Rmd in advance and including it in the UI function instead.

Locally, the code as proposed in the PR performs faster than the solution using `includeHTML("https://sage-bionetworks.github.io/dccvalidator/articles/using-the-dccvalidator-app-amp-ad.html")`.  I imagine we would make this an optional configuration to not burden an end user with making their own pkgdown site? 

> Include the vignettes in the inst/ directory so that they will be installed when the user installs dccvalidator. Then we could access them with system.file() the way we currently do for the CSS in the UI:

These vignettes are not designed to be extensible in their current format therefore I don't see them retaining much utility for a user installing dccvalidator. 

I guess I am leaning towards option 1, though I would need further information on making that into an optional configuration and a deeper look at speed. 

> Would including a link to the pkgdown site be an ok short-term measure if we can't get the embedded page fully worked out by then?

I can implement this change temporarily on my fork.

karawoo:
> Locally, the code as proposed in the PR performs faster than the solution using includeHTML

We would have to test this on the server -- when running it locally, you're initializing the app and connecting to it at the same time. In production, these happen at different times. We initialize the app with `touch restart.txt` and at that point the UI function gets run. This generates an HTML file which is cached. When a user connects to the app the UI already exists, and only the server function gets run. It's possible that including the HTML could make starting the app slower but connecting to it faster.

>  I imagine we would make this an optional configuration to not burden an end user with making their own pkgdown site

That's what I imagined -- the URL could be in the config and it could point to our pkgdown site by default or could be overridden to point to any HTML file.

> These vignettes are not designed to be extensible in their current format therefore I don't see them retaining much utility for a user installing dccvalidator.

In that case maybe we should make it the default behavior to not include vignettes in the app at all, and allow ourselves to override that somehow.

Let me know if you want to sit down and pair on this when you're back from vacation.

kelshmo:
I was not able to strip out the pkgdown header so I am testing option 2 - Include the vignettes in the `inst/` directory.

kelshmo:
@karawoo I tested this locally with `R_CONFIG_ACTIVE = pec` and looks to be working well! If you approve of this model, I can push the changes for AMP-AD as well.

R/app-server.R
```diff
@@ -9,6 +9,18 @@
...
+    HTML(markdown::markdownToHTML(
+      knitr::knit(input = glue::glue(
+        tools::file_path_sans_ext(config::get("path_to_markdown")), ".Rmd"
+        ),
+        output = glue::glue(
+          tools::file_path_sans_ext(config::get("path_to_markdown")), ".md"
+        ),
+                  quiet = TRUE),
+      stylesheet = ("inst/app/www/custom.css")))
+  })
```
Aryllen: I thought I left this comment a while ago, but it seems to not be here.

One of the issues mentioned was that the markdown takes time to render and, since it doesn't rely on anything needed from the user, could be done in `app_ui()` to decrease user wait time. To do so, this would be removed from the server function and put into a separate function (perhaps in app-ui.R) without the renderUI call: https://github.com/kelshmo/dccvalidator/blob/0f947f5b8b21328c786d6e1239574fd5629a644b/R/app-server.R#L12-L23

New function example:

```
get_markdown <- function() {
    HTML(markdown::markdownToHTML(
        knitr::knit(input = glue::glue(
            tools::file_path_sans_ext(config::get("path_to_markdown")), ".Rmd"
        ),
        output = glue::glue(
            tools::file_path_sans_ext(config::get("path_to_markdown")), ".md"
        ),
        quiet = TRUE)
}
```

Then the `app_ui()` would call the new function to generate the markdown and get the html.
```
tabItem(
          tabName = "vignette",
          get_markdown()
        ),
```

I believe this should work and would put the rendering burden on the initial spin-up of the app, as opposed to re-rendering for every user.

R/app-server.R
```diff
@@ -9,6 +9,18 @@
...
+  output$markdown <- renderUI({
+    HTML(markdown::markdownToHTML(
+      knitr::knit(input = glue::glue(
+        tools::file_path_sans_ext(config::get("path_to_markdown")), ".Rmd"
+        ),
+        output = glue::glue(
+          tools::file_path_sans_ext(config::get("path_to_markdown")), ".md"
+        ),
+                  quiet = TRUE),
+      stylesheet = ("inst/app/www/custom.css")))
```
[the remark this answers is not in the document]
kelshmo: When I remove this line, the app does not scale to screen.
Screen Shot 2019-12-26 at 12 04 48 PM

R/app-server.R
kelshmo: This works locally, thanks @Aryllen ! @karawoo what do you think about implementing this change?

kelshmo:
@karawoo I tested adding the `vignette` header information to the markdown in the `vignettes/` folder, adding `system.file()` in the code chunk and then installing from the branch `devtools::install_github("kelshmo/dccvalidator@link-vignette", build_vignettes = TRUE, force = TRUE)`. I return this error: 
`Warning in file(con, "r") :
  cannot open file 'inst/using-the-dccvalidator-pec.Rmd': No such file or directory`

I read around a bit on `inst/` and this folder should install at the top directory. Do you know how to test that this folder was installed? Can you repro this error?

R/app-server.R
karawoo: Yeah that sounds great!

karawoo:
> Do you know how to test that this folder was installed?

If you do `system.file("using-the-dccvalidator-pec.Rmd", package = "dccvalidator")` in the R console it should print the path to where the file is located. If the file wasn't installed, then it should return an empty string.

> Can you repro this error?

No, when I do `devtools::install_github("kelshmo/dccvalidator@link-vignette", build_vignettes = TRUE, force = TRUE)` it works, and then I can see the PEC vignette when I do `browseVignettes("dccvalidator")`.

> Also, we need to address why the app does not scale to the full screen when the CSS is removed from HTML() in app-server.R - Seems surprising it would have this effect.

Yeah I'm not sure what that's about. Does that still happen if you move the code to the UI as @Aryllen suggested?

kelshmo:
I moved the function to `app-ui.R` with the most recent commit. 

- When I run the app locally from my fork, it does not scale to fullscreen. When I run the app from installation, it looks like it does scale appropriately but I still return an error on the "Using the App" page. My system can find the file but the path is not nested in the `/inst` folder. 
> system.file("using-the-dccvalidator-pec.Rmd", package = "dccvalidator")
[1] "/Users/kmontgomery/Library/R/3.6/library/dccvalidator/using-the-dccvalidator-pec.Rmd"

I'm puzzled.

karawoo:
> My system can find the file but the path is not nested in the /inst folder.

That's correct, the contents of `inst/` folder get installed at the top level of the package directory, but the `inst/` folder itself doesn't. Basically, everything that's within `inst/` gets moved up a directory.

I'm not sure why you're seeing this behavior with the errors and formatting, can we sit down in person or on a call to walk through it next week?

R/mod-app-instructions.R
```diff
@@ -0,0 +1,12 @@
+get_markdown <- function() {
+  HTML(markdown::markdownToHTML(
+    knitr::knit(input = glue::glue(
+      tools::file_path_sans_ext(config::get("path_to_markdown")), ".Rmd"
+    ),
+    output = glue::glue(
+      tools::file_path_sans_ext(config::get("path_to_markdown")), ".md"
+    ),
+    quiet = TRUE)
```
karawoo:
```suggestion
    quiet = TRUE), 
    stylesheet = system.file("app/www/custom.css", package = "dccvalidator")
```

kelshmo:
@karawoo It looks to me like the following config option does not work with a fork of the repo: 
`path_to_markdown: !expr system.file("using-the-dccvalidator-pec.Rmd", package = "dccvalidator")`
When I execute the following, I return an empty string and thus an error in starting the app: 
```
> config::get("path_to_markdown")
[1] ""
```

R/app-ui.R
```diff
@@ -12,6 +12,9 @@ app_ui <- function(request) {
 
     dashboardSidebar(
       sidebarMenu(
+        conditionalPanel(!is.na(config::get("path_to_markdown")) == TRUE, {
+            menuItem("Using the App", tabName = "vignette")
+        }),
```
kelshmo: @Aryllen Do you know how to add a conditional `menuItem()`? I googled around and tested some `if statements` to no avail. In the case of the code as written, the menu item is not clickable but the text "Using the App" is still displayed.

kelshmo: I should add I was testing this with the `config.yml` set to `path_to_markdown: !expr NA`

karawoo:
> It looks to me like the following config option does not work with a fork of the repo:
path_to_markdown: !expr system.file("using-the-dccvalidator-pec.Rmd", package = "dccvalidator")

When running it within your fork have you installed an updated version of dccvalidator from your branch? R is trying to find the file from the location where dccvalidator is installed. If you're using the current master branch version of dccvalidator, this file will not be there. `devtools::install()` (and restart R) should fix it locally, I think.

R/app-ui.R
Aryllen: I'm not sure a `conditionalPanel` would work in that case. What you could try is to make a function that checks if the `menuItem` should be created. If so, it gives back the code generated by `menuItem("Using the App", tabName = "vignette")`. If not, it returns nothing.

karawoo: @kelshmo if it's too hard to implement this let's not worry too much about it and just choose one of the vignettes to make the default.

Aryllen: Also, this is just a heads up for the future since I often forget myself, but the conditionalPanel needs the condition in javascript ([documentation](https://shiny.rstudio.com/reference/shiny/0.11/conditionalPanel.html)). I think a separate function with an if-else _should_ work, however, and be simpler than conditionalPanels.

kelshmo: I swear I tried a good old `if` statement and it was still displaying the text in the browser. Now, it appears to work well...!

kelshmo:
I believe this is complete. 

Issue #246 will take care of documenting the use of NA to ignore the `path_to_markdown` config option.

config.yml
```diff
@@ -71,6 +72,7 @@ amp-ad:
 
 pec:
   parent: "syn20825282"
+  path_to_markdown: !expr system.file("using-the-dccvalidator-pec.Rmd", package = "dccvalidator")
```
karawoo:
```suggestion
  path_to_markdown: "inst/using-the-dccvalidator-pec.Rmd"
```

config.yml
```diff
@@ -1,5 +1,6 @@
 default:
   parent: "syn20400157"
+  path_to_markdown: !expr system.file("using-the-dccvalidator-app-amp-ad.Rmd", package = "dccvalidator")
```
karawoo:
```suggestion
    path_to_markdown: "inst/using-the-dccvalidator-app-amp-ad.Rmd"
```
I know I suggested this, but it is not working when I run the app from my fork. After I do `load_all()`, `system.file("using-the-dccvalidator-app-amp-ad.Rmd", package = "dccvalidator")` points to the local copy of the Rmd, not the installed one, which is fine but for some reason `config::get("path_to_markdown")` only returns an empty string. 

I suggest that for our copy of the config file, we use the local copy of the Rmd. If someone installs the package without forking and wants to use our documentation, they could still do that with `system.file()`. They wouldn't run into the issue above since they wouldn't be using `load_all()`.

vignettes/customizing-dccvalidator.Rmd
```diff
@@ -63,6 +63,10 @@ default:
 ## Configuration options
 
 * `parent`: The Synapse project or folder where files will be stored
+* `path_to_markdown`: The `system.file()` path to the markdown document with app 
+  instructions. This option must follow the format 
+  `!expr system.file("path_to_md_document.Rmd", package = "dccvalidator")`. 
+  If you wish to omit app embedded instructions, insert `!expr NA`.
```
karawoo: I suggest something like "Location of an R Markdown document with app instructions. If you wish to omit instructions, insert `!expr NA`."

config.yml
kelshmo: At what point does a user execute `load_all()`? I only familiar with deploying the app using the [instructions](https://sage-bionetworks.github.io/dccvalidator/articles/deploying-dccvalidator.html).

karawoo: It's in the `app.R` file, so it gets run if you deploy the app by forking & cloning the repo

kelshmo: Ok, got it!

PR #200, opened 2019-11-14
Add link and text for study acknowledgments to document upload

### LINKED ISSUE #198: Add acknowledgement statement text
@amapeters requested to add some text about how contributors can upload an acknowledgement preference in the study documentation. Text to add in bold:

> # Upload Study Documentation
> This information should be similar to a materials and methods section in a paper. An example of what a study should include can be found here for an animal model study and here for a human study. **If you wish, also provide an acknowledgement statement and/or reference that should be included in publications resulting from secondary data use, examples can be found here: https://adknowledgeportal.synapse.org/#/DataAccess/AcknowledgementStatements. This can be provided as part of the study documentation text**

Aryllen:
Fixes #198.

This update adds:
- text in study documentation instructions for what acknowledgments are,
- example of acknowledgements via link passed into the function,
- config addition for customizing the example link

@kelshmo, the config will need an example to work for your app.


karawoo:
@Aryllen I was talking with @amapeters earlier and this text will not be relevant to PsychENCODE because they have one acknowledgment statement that is used for everything. Could this be reworked to enable/disable the text altogether based on the config? Either the full text could be in the config, or there could be a function that generates it that we call based on the config?

Aryllen:
Sure! Putting the full text in the config seems like a lot to put in there (including some code bits), but I will think about how to do that as a function based on the config and update it.

Aryllen:
Updated with a function version.

R/mod-upload-documentation.R
```diff
@@ -202,3 +140,106 @@ upload_documents_server <- function(input, output, session,
...
+      ),
+      tags$li(
+        "cell type,"
+      ),
+      tags$li(
+        "cell culture information (such as primary or immortalized cell line, passage, treatments, differentiation). If a commercial cell line, provide a link."
+      )
+    ),
+
+    p("* Include citations for more study information of available."),
```
karawoo:
```suggestion
    p("Include citations for more study information if available."),
```

R/mod-upload-documentation.R
```diff
@@ -202,3 +140,106 @@ upload_documents_server <- function(input, output, session,
...
+                                      study_link_ref) {
+  if (study_link_ref != "") {
+    overview_text <- p(
+      # nolint start
+      "This information should be similar to a materials and methods section in a paper. An example of what a study should include can be found ",
+      HTML(glue::glue("<a target =\"_blank\" href=\"{study_link_animal}\">here</a> for an animal model study and ")),
+      HTML(glue::glue("<a target =\"_blank\" href=\"{study_link_human}\">here</a> for a human study.")),
+      "If you wish, also provide an acknowledgement statment and/or reference that should be included in publications resulting from secondary data use; examples can be found ",
+      HTML(glue::glue("<a target =\"_blank\" href=\"{study_link_ref}\">here</a>.")),
+      "This can be provided as part of the study or assay documentation text."
```
karawoo: To simplify things I suggest we tell people to provide it with the study description:

```suggestion
      "This can be provided as part of the study documentation text."
```

R/mod-upload-documentation.R
Aryllen: Fix grammar error.

R/mod-upload-documentation.R
Aryllen: Make consistent with instructions.

Aryllen:
Fixed. Thanks!

karawoo:
whoops @Aryllen I don't think you actually committed the changes

Aryllen:
Haha, whoops. Comments don't do a whole lot to fix stuff. XD Now they should be fixed, @karawoo.

PR #201, opened 2019-11-14
Remove name parameter from `save_to_synapse()` & update calls

### LINKED ISSUE #199: Remove filename from save_to_synapse() function signature
I think we should change this function signature a little bit. It doesn't seem right to have a default value that we then say is invalid (even though in practice we usually override the default value). And the reason `name` defaults to `NULL` in `save_to_synapse()` was because I thought that was the default value of `File()`, but it turns out when you explicitly pass `NULL` to `File()` you change the behavior: https://sagebionetworks.jira.com/browse/SYNR-1437.

What do you think about removing `name` as an argument altogether, since we seem to always want to get it from the input file anyway?

```r
save_to_synapse <- function(input_file, parent, annotations = NULL) {
  file_to_upload <- synapser::File(
    input_file$datapath,
    parent = parent,
    name = input_file$name,
    annotations = annotations
  )
  if (!is_name_valid(file_to_upload$properties$name)) {
    stop("Please check that file names only contains: letters, numbers, spaces, underscores, hyphens, periods, plus signs, and parentheses.") # nolint
  }
  synapser::synStore(file_to_upload)
}
```

_Originally posted by @karawoo in https://github.com/Sage-Bionetworks/dccvalidator/pull/191_

Aryllen:
Fixes #199.

The `input_file` parameter includes the file name already. This update:
- removes the redundant `name` parameter from `save_to_synapse()`,
- updates the synapse file entity call to use `input_file$name`,
- updates all calls to `save_to_synapse()` to remove the name parameter.

### DIFF
--- a/R/app-server.R
+++ b/R/app-server.R
@@ -400,7 +400,6 @@ app_server <- function(input, output, session) {
           save_to_synapse(
             files$indiv,
             parent = created_folder,
-            name = files$indiv$name,
             annotations = list(
               study = study_name(),
               metadataType = "individual",
@@ -412,7 +411,6 @@ app_server <- function(input, output, session) {
           save_to_synapse(
             files$biosp,
             parent = created_folder,
-            name = files$biosp$name,
             annotations = list(
               study = study_name(),
               metadataType = "biospecimen",
@@ -424,7 +422,6 @@ app_server <- function(input, output, session) {
           save_to_synapse(
             files$assay,
             parent = created_folder,
-            name = files$assay$name,
             annotations = list(
               study = study_name(),
               metadataType = "assay",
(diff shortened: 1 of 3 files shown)


Aryllen:
I had that same thought about 5 minutes after I asked for your feedback on the idea this morning. Haha. I checked it out and it doesn't affect the appending of the study name. `save_to_synapse()` expects `input_file` to be a list. I send in a list with the datapath and the new file name for study docs. 

https://github.com/Sage-Bionetworks/dccvalidator/blob/ded4f9f465d7f4388a006a9abbf0f2e0e77e02ba/R/mod-upload-documentation.R#L188-L198

PR #202, opened 2019-11-14
Update check_ids_match reporting based on bidirectional flag

### LINKED ISSUE #196: When checking IDs in one direction, errors are reported for both
Screen Shot 2019-11-12 at 3 45 41 PM

Only "missing from individual" should be shown here since we are checking with `bidirectional = FALSE` (#190)

### LINKED ISSUE #197: Deal with concurrent updates to master metadata files
Scenario: Person A and person B are both updating study X. They both add different specimens to the master biospecimen file (or individuals to the individual file). If they upload this file to the same staging folder, one of two undesirable things will happen: 1) if the file names are the same, they will overwrite one another, 2) if the file names are not the same, we will still need to manually combine the two files to have a complete record of all the specimens. Not sure how to combat this from the validator, but we need to think about it...

Aryllen:
Fixes #196.

Updates the `check_ids_match()` to only return relevant missing values based on the bidirectional flag. Added test to check that the reporting is correct based on the flag.

### DIFF
--- a/tests/testthat/test-check-ids-match.R
+++ b/tests/testthat/test-check-ids-match.R
@@ -185,7 +185,8 @@ test_that("check_ids_match bidirectional arg returns relevant data", {
   )
 expect_equal(res1$data, list(`Missing from x` = 5, `Missing from y` = 3:4))
 expect_equal(length(res1$data), 2)
-  expect_equal(res2$data[[1]], 5)
+  expect_equal(res2$data, list(`Missing from x` = 5))
+  expect_equal(length(res2$data), 1)
 })
 
 test_that("check_ids_match data gets default names if not provided", {
--- a/tests/testthat/test-check-ids-match.R
+++ b/tests/testthat/test-check-ids-match.R
@@ -183,7 +183,8 @@ test_that("check_ids_match bidirectional arg returns relevant data", {
     idcol = "individualID",
     bidirectional = FALSE
   )
-  expect_equal(c(res1$data[[1]], res1$data[[2]]), c(5, 3, 4))
+expect_equal(res1$data, list(`Missing from x` = 5, `Missing from y` = 3:4))
+expect_equal(length(res1$data), 2)
   expect_equal(res2$data[[1]], 5)
 })
 
--- a/tests/testthat/test-check-ids-match.R
+++ b/tests/testthat/test-check-ids-match.R
@@ -183,8 +183,11 @@ test_that("check_ids_match bidirectional arg returns relevant data", {
     idcol = "individualID",
     bidirectional = FALSE
   )
-expect_equal(res1$data, list(`Missing from x` = 5, `Missing from y` = 3:4))
-expect_equal(length(res1$data), 2)
+  expect_equal(
+    res1$data,
+    list(`Missing from x` = 5, `Missing from y` = 3:4)
+  )
+  expect_equal(length(res1$data), 2)
   expect_equal(res2$data, list(`Missing from x` = 5))
   expect_equal(length(res2$data), 1)
 })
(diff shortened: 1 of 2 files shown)


tests/testthat/test-check-ids-match.R
```diff
@@ -168,6 +168,25 @@ test_that("check_ids_match bidirectional arg looks only in one direction", {
...
+    idcol = "individualID",
+    bidirectional = TRUE
+  )
+  res2 <- check_ids_match(
+    x = meta,
+    y = manifest,
+    idcol = "individualID",
+    bidirectional = FALSE
+  )
+  expect_equal(c(res1$data[[1]], res1$data[[2]]), c(5, 3, 4))
```
karawoo:
```suggestion
expect_equal(res1$data, list(`Missing from x` = 5, `Missing from y` = 3:4))
expect_equal(length(res1$data), 2)
```
It took me a minute to interpret what this test was doing, can we update it to compare the exact data that is being returned?

tests/testthat/test-check-ids-match.R
```diff
@@ -168,6 +168,25 @@ test_that("check_ids_match bidirectional arg looks only in one direction", {
...
+  expect_equal(res2$data[[1]], 5)
```
karawoo: This one doesn't test the new behavior that we've added, as it would pass even if both sets of values were being returned. I'd propose something like:
```suggestion
  expect_equal(res2$data, list(`Missing from x` = 5))
  expect_equal(length(res2$data), 1)
```

tests/testthat/test-check-ids-match.R
Aryllen: Sure! That definitely helps readability.

tests/testthat/test-check-ids-match.R
Aryllen: Oh! I see. That makes sense. Whoops. We need to check that it's not returning both. Thank you!

tests/testthat/test-check-ids-match.R
```diff
@@ -168,6 +168,27 @@ test_that("check_ids_match bidirectional arg looks only in one direction", {
...
+  )
+  res2 <- check_ids_match(
+    x = meta,
+    y = manifest,
+    idcol = "individualID",
+    bidirectional = FALSE
+  )
+expect_equal(res1$data, list(`Missing from x` = 5, `Missing from y` = 3:4))
+expect_equal(length(res1$data), 2)
+  expect_equal(res2$data, list(`Missing from x` = 5))
```
karawoo:
```suggestion
 expect_equal(res2$data, list(`Missing from x` = 5))
```

tests/testthat/test-check-ids-match.R
```diff
@@ -168,6 +168,27 @@ test_that("check_ids_match bidirectional arg looks only in one direction", {
...
+  expect_equal(length(res2$data), 1)
```
karawoo:
```suggestion
 expect_equal(length(res2$data), 1)
```

tests/testthat/test-check-ids-match.R
```diff
@@ -168,6 +168,27 @@ test_that("check_ids_match bidirectional arg looks only in one direction", {
...
+    idcol = "individualID",
+    bidirectional = TRUE
+  )
+  res2 <- check_ids_match(
+    x = meta,
+    y = manifest,
+    idcol = "individualID",
+    bidirectional = FALSE
+  )
+expect_equal(res1$data, list(`Missing from x` = 5, `Missing from y` = 3:4))
```
karawoo:
```suggestion
 expect_equal(res1$data, list(`Missing from x` = 5, `Missing from y` = 3:4))
```

tests/testthat/test-check-ids-match.R
```diff
@@ -168,6 +168,27 @@ test_that("check_ids_match bidirectional arg looks only in one direction", {
...
+expect_equal(length(res1$data), 2)
```
karawoo:
```suggestion
  expect_equal(length(res1$data), 2)
```

Aryllen:
That is weird. I changed the spacing locally and pushed. There would have been a lint problem with one of them so I fixed that, as well.

PR #203, opened 2019-11-14
Add descriptive messages to manifest annotations key & value checks

### LINKED ISSUE #195: When checking annotation values, include "manifest" in resulting message
e.g. "Some annotation keys are invalid" should be "Some annotation keys are invalid in the manifest"

Aryllen:
Fixes #195.

This update passes in success/fail messages to the manifest key and value annotation checks.

### DIFF
--- a/R/app-server.R
+++ b/R/app-server.R
@@ -268,7 +268,7 @@ app_server <- function(input, output, session) {
         manifest(),
         annots,
         whitelist_keys = c("path", "parent"),
-        success_msg = "All values in the manifest are valid",
+        success_msg = "All keys (column names) in the manifest are valid",
         fail_msg = "Some values in the manifest are invalid"
       )
     })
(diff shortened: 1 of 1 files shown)


R/app-server.R
```diff
@@ -267,13 +267,20 @@ app_server <- function(input, output, session) {
       check_annotation_keys(
         manifest(),
         annots,
-        whitelist_keys = c("path", "parent")
+        whitelist_keys = c("path", "parent"),
+        success_msg = "All values in the manifest are valid",
```
```suggestion
        success_msg = "All keys (column names) in the manifest are valid",
```

R/app-server.R
```diff
@@ -267,13 +267,20 @@ app_server <- function(input, output, session) {
...
+        fail_msg = "Some values in the manifest are invalid"
```
```suggestion
        fail_msg = "Some keys (column names) in the manifest are invalid"
```

PR #214, opened 2019-11-21
Check that metadata files are included in manifest

### LINKED ISSUE #75: Metadata files should be present in manifest
I've run into this working with collaborators in the past: they don't include the metadata files in the manifest, which sometimes causes confusion and we've needed to go back and track down the files or upload them ourselves from email attachments. We could add a check that all of the metadata files they've uploaded to the app are also present in the `path` column of the manifest.

karawoo:
Fixes #75. Checks that any metadata files that the user has uploaded to the Shiny app for validation are also included in the `path` column of the manifest.


karawoo:
That's a good catch. I think it should be easy to fix. If the filenames are all NULL perhaps we should just return NULL. Then if someone validates the manifest and does not validate any metadata files, they won't see a result for this check at all (which seems appropriate). How does that sound to you?

Aryllen:
I think that would work well for the situation that the user validates the manifest by itself. 

It would not help in the case that the user validates the manifest and any subset of the metadata files. For example, manifest includes individual.csv, but not biospecimen.csv or assay.csv, but it _should_ include all of them for that particular dataset. 
- Currently, if the user validates the manifest alone, the check is successful.
- If the user validates manifest + individual, the check is successful. 
- If the user later validates manifest + individual + biospecimen (or manifest + biospecimen), the check fails. Similarly, the check fails for validating manifest + assay, and all other combinations of files that are not the two successful checks.

This works in both directions, they could try to validate a failing combination, but then try a succeeding combination. They may be confused as to why it failed the first time. Basically, the inconsistency could be confusing. Additionally, I'm not sure what we are requesting for validation success. If it is a screenshot, for example, does it need to include all files at once? If so, this would limit confusion since the final result, every time, is a fail until the manifest gets the metadata files included.

In short, that would solve the one problem where the manifest is checked alone, but there are many situations in which it would still pop up. I don't think it's such a large problem that all cases would need to be accounted for, though. The confusion can be solved with a logical explanation.

karawoo:
Ah yes I see what you mean. I don't see a good way around that one because we don't know the file names for their metadata in advance, we only know the names of whatever files they upload to the validator. I don't think it's _too_ confusing though since the details show which file is missing. If their manifest contains only the individual file and they validate manifest + individual, then later manifest + individual + biospecimen, they can see that it's the biospecimen file that is missing. I think this is probably ok?

Aryllen:
Yup, I agree. Thank you for adding the feature to make it return `NULL` if there are no other files uploaded, as well.

PR #219, opened 2019-11-22
More thorough documentation about how to spin up a copy of the app

karawoo:
- Adds alternate method of running the app to the customization vignette
- Points to customization vignette instructions from readme

### DIFF
--- a/vignettes/customizing-dccvalidator.Rmd
+++ b/vignettes/customizing-dccvalidator.Rmd
@@ -40,6 +40,17 @@ To create a custom version of the app, you'll need to follow these steps:
    [Deploying dccvalidator](https://sage-bionetworks.github.io/dccvalidator/articles/deploying-dccvalidator.html) 
    vignette.
 
+Alternatively, instead of forking the repository you can install dccvalidator
+and create an `app.R` file containing the following:
+
+```
+library("dccvalidator")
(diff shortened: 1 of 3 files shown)


vignettes/customizing-dccvalidator.Rmd
```diff
@@ -40,6 +40,17 @@ To create a custom version of the app, you'll need to follow these steps:
    [Deploying dccvalidator](https://sage-bionetworks.github.io/dccvalidator/articles/deploying-dccvalidator.html) 
    vignette.
 
+Alternatively, instead of forking the repository you can install dccvalidator
+and create an `app.R` file containing the following:
+
+```
+library("dccvalidator")
+run_app()
```
Aryllen: This would use "default" as the config, correct? If so, we may want to specify that they should make their config be the default, as opposed to naming it as the instructions (instruction 5).

karawoo: Yes -- ultimately I'd like to make the configuration file, and perhaps which configuration to use, a parameter to to the app, but for now it depends on a config file in the same directory with a default config.

Aryllen: Alrighty. Sounds good.

PR #227, opened 2019-12-05
Replace synapser with reticulate

### LINKED ISSUE #155: 403 Error on repeat submission
There's an issue with the sessions in that some information must be being shared. When repeatedly submitting the same file, sometimes a 403 Error appears where it's thinking the file belongs to someone else.

image

### LINKED ISSUE #216: Replace synapser with reticulate + python client
This is one solution to #155 and would also enable us to support the specific JSON schema validation package already being used elsewhere for validation (#72).

karawoo:
Fixes #216, fixes #155. This is a draft PR because I still need to add tests for some functions.

- Replaces all uses of synapser with reticulate + synapse Python client, and handles/converts new data structures and classes when necessary
- Adds new parameters to functions so that synapse client objects can be passed in, and provides those parameters when functions are called
- Reimplements functions in dependencies that use synapser (e.g. syndccutils) and removes those dependencies
- Installs synapse client on Travis for testing

- Added tests for new functions or for new behavior in existing functions: not yet


Aryllen:
Could you please reinitialize the renv.lock file? It still has PythonEmbedInR.

karawoo:
I have deployed this branch at https://shinypro.synapse.org/users/kwoo/dccvalidator-staging/; @Aryllen @kelshmo since this is such a big change I would appreciate it if you could kick the tires on this extra thoroughly.

Aryllen:
> I have deployed this branch at https://shinypro.synapse.org/users/kwoo/dccvalidator-staging/; @Aryllen @kelshmo since this is such a big change I would appreciate it if you could kick the tires on this extra thoroughly.

Great! Will do.

karawoo:
I have deployed this branch at https://shinypro.synapse.org/users/kwoo/dccvalidator-staging/; @Aryllen @kelshmo since this is such a big change I would appreciate it if you could kick the tires on this extra thoroughly.

Aryllen:
Just tried it out and it seems like I'm getting more errors than usual. It's saying things like individualID, specimenID, specimenIDSource are invalid values. My first assumption is that these are being treated as factors instead of strings, but I'm not sure why changing to reticulate would affect that. So far, everything else seems to be working fine.

karawoo:
Aha, good catch. I think I've fixed that.

Aryllen:
> Aha, good catch. I think I've fixed that.

Great! I haven't found any other problems. It seems to be working. Of course, I can't test the user bug alone so not sure on that part.

karawoo:
Yeah, let's try that together next week. I might need to deploy it in a different location where the shiny server utilization scheduler hasn't been customized so we can be sure this change solves the issue on its own.

karawoo:
Tested this on the shiny server with @Aryllen: we created a new app (in a location that does not have a customized utilization scheduler). I connected and uploaded a file to the validator, then Nicole connected, then I uploaded the same file again. No error occurred, and the shiny admin console showed us both connected to the same R process. So I think we can confirm that this solution does indeed solve #155.

karawoo:
I've updated the app on https://shinypro.synapse.org/users/kwoo/dccvalidator-staging/ with the latest changes. From my testing, things seem to be working. I think this is ready for review 🤞

kelshmo:
While testing locally, I use `renv::restore()` to load package dependencies and return this error when I try to run the app `Error in python_binary_path(reticulate_python_env) : 
  failed to discover Python binary associated with path '/Users/kmontgomery/Documents/dccvalidator/renv/python/r-reticulate'`. I will test on the Shiny server next.

karawoo:
Hmm interesting, I can't reproduce that even with a fresh clone of the repo. Do you get the error when calling `run_app()`, or `shiny::runApp()`/clicking the run app button in RStudio? What is the output of `reticulate::py_config()`?

kelshmo:
`reticular::pyconfig()` returns the same error. Interestingly, `run_app()` does not trigger the error, only `shiny::runApp()` or selecting the run app button does.

karawoo:
Fascinating. Can you confirm you have a recent checkout of this branch? At one point I had included a python environment within the repo, but took that out due to some python version issues that I have an open jira about. So right now, `runApp()` or clicking the RStudio button should source the app.R file, which doesn't do anything python-specific that I'm aware of 🤔

kelshmo:
I can confirm I'm on the most recent checkout of this branch.

karawoo:
Can you share the output of `Sys.getenv("RETICULATE_PYTHON")`?

kelshmo:
It is an empty list.
```
> Sys.getenv("RETICULATE_PYTHON")
[1] ""
```

karawoo:
Ok, that's expected. Could you also share the output of `traceback()` if you run it right after you hit the error?

kelshmo:
```
13: execCallbacks(timeoutSecs, all, loop$id)
12: force(expr)
11: with_loop(loop, invisible(execCallbacks(timeoutSecs, all, loop$id)))
10: run_now(timeoutMs/1000, all = FALSE)
9: service(timeout)
8: serviceApp()
7: ..stacktracefloor..(serviceApp())
6: withCallingHandlers(expr, error = doCaptureStack)
5: domain$wrapSync(expr)
4: promises::with_promise_domain(createStackTracePromiseDomain(), 
       expr)
3: captureStackTraces({
       while (!.globals$stopped) {
           ..stacktracefloor..(serviceApp())
       }
   })
2: ..stacktraceoff..(captureStackTraces({
       while (!.globals$stopped) {
           ..stacktracefloor..(serviceApp())
       }
   }))
1: runApp()
```

kelshmo:
It works!! 
- to document: `synapseclient` must be installed on the python version that `reticulate` points to in order for the app to function. For me, that was python3. The fix was to:
1. restart R
2. `library("reticulate")`
3. `use_python("/usr/local/bin/python3", required = TRUE)`

PR #232, opened 2019-12-10
Update documentation

karawoo:

- Adds package-level documentation (viewable with `?dccvalidator`), including handy URLs to the pkgdown site and GitHub


Aryllen:
Oh, I did just notice that the tab for the online documentation has CNS in it. Not sure if you want to keep it like that.

image

karawoo:
The site won't be updated until this PR is merged, but I confirmed that when I build it locally "CNS" no longer appears 🙂

Aryllen:
> The site won't be updated until this PR is merged, but I confirmed that when I build it locally "CNS" no longer appears 🙂

Nice!

Issue #236, opened 2019-12-20
Intermittently failing test on travis


karawoo:
[This test](https://github.com/Sage-Bionetworks/dccvalidator/commit/26ba9fc788065de45537c699c44b782041f39ee5) fails just some of the time with the following info on travis: 

```
 1. Error: get_synapse_table reads in table with empty strings as NA (@test-ge

SynapseHTTP400 Client 

The passed reset-token was invalid. The table's status was reset after the passed reset-token was acquired.

Detailed traceback: 

  File "/home/travis/.local/lib/python2.7/site-packages/synapseclient/client.py", line 2875, in tableQuery

    return CsvFileTable.from_table_query(self, query, **kwargs)

  File "/home/travis/.local/lib/python2.7/site-packages/synapseclient/table.py", line 1601, in from_table_query

    includeRowIdAndRowVersion=includeRowIdAndRowVersion)

  File "/home/travis/.local/lib/python2.7/site-packages/synapseclient/client.py", line 3033, in _queryTableCsv

    download_from_table_result = self._waitForAsync(uri=uri, request=download_from_table_request)

  File "/home/travis/.local/lib/python2.7/site-packages/synapseclient/client.py", line 2740, in _waitForAsync

    result = self.restGET(uri+'/get/%s' % async_job_id['token'], endpoint=endpoint)

  File "/home/travis/.local/lib/python2.7/site-packages/synapseclient/client.py", line 3449, in restGET

    exceptions._raise_for_status(response, verbose=self.debug)

  File "/home/travis/.local/lib/python2.7/site-packages/synapseclient/exceptions.py", line 153, in _raise_for_status

    raise SynapseHTTPError(message, response=response)

Backtrace:

 1. dccvalidator::get_synapse_table(synID, syn = syn)

 2. syn$tableQuery(glue::glue("select * from {synID}"), includeRowIdAndRowVersion = FALSE)

 3. reticulate:::py_call_impl(callable, dots$args, dots$keywords)
```

No idea what this error means but I don't think it's a problem with the `get_synapse_table()` function. Maybe we can fix it by using a pre-existing table instead of creating one in the test.

PR #238, opened 2019-12-20
Remove observeEvent assignment

karawoo:
This was copied from the [SynapseShinyApp template](https://github.com/Sage-Bionetworks/synapseshinyapp) and isn't necessary as far as I can tell.


karawoo:
Failing test is just due to #236, should be fixed in #237

PR #242, opened 2019-12-26
split instructions

### LINKED ISSUE #241: Clarify instructions to set up app from dccvalidator
I'm reading through [these instructions](https://sage-bionetworks.github.io/dccvalidator/articles/customizing-dccvalidator.html) to try to test installing the app from the `dccvalidator` package.

I suggest splitting the instructions into two options 1. Install from a fork or 2. Install the package. As it reads now, the instructions are difficult to follow if one wants to install from the package. For example, one needs to follow step 1 and 3 but not 2 if installing the `dccvalidator`. Also, does a user still need to create an `app.R` file? I was able to run the app by executing `dccvalidator::app_run()`.

Additionally, I believe that a `config.yml` file requires a default configuration so that could also be noted here.

kelshmo:
Fixes #241

### DIFF
--- a/vignettes/customizing-dccvalidator.Rmd
+++ b/vignettes/customizing-dccvalidator.Rmd
@@ -30,7 +30,7 @@ To create a custom version of the app, you'll need to follow these steps:
 1. Fork the [dccvalidator](https://github.com/Sage-Bionetworks/dccvalidator/blob/master/config.yml)
    GitHub repository.
 1. Create a new configuration in the `config.yml` file. Note that any values you
-   do not customize will be inherited from the default configuration. YML format requires a default configuration.
+   do not customize will be inherited from the default configuration. The configuration file must have a `default` configuration.
 1. (Optional): create a pull request with your configuration back to the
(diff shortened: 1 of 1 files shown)


vignettes/customizing-dccvalidator.Rmd
```diff
@@ -30,7 +30,7 @@ To create a custom version of the app, you'll need to follow these steps:
 1. Fork the [dccvalidator](https://github.com/Sage-Bionetworks/dccvalidator/blob/master/config.yml)
    GitHub repository.
 1. Create a new configuration in the `config.yml` file. Note that any values you
-   do not customize will be inherited from the default configuration.
+   do not customize will be inherited from the default configuration. YML format requires a default configuration.
```
karawoo: Can we tweak the wording of this slightly? It's the config R package that requires the default configuration (not anything to do with the yaml format) but that's probably more detail than is needed. Since the config.yml file already has the default section, maybe we can say something like "The configuration file must have a `default` configuration."

Issue #217, opened 2019-11-21
Submit to CRAN


karawoo:
The package isn't suitable for submission yet but I would like to submit eventually. A few things I can think of will need to be changed, so I'm starting this issue to keep track of them.

- [x] Remove synapser (#216)
- [x] Examples for all exported functions
- [x] Reduce screenshot size (package should be <5MB total)
- [x] Return value documentation for all exported functions
- [x] Figure out if examples should be in `\dontrun{}` or `\donttest{}` -- there's mixed info out there on this
- [x] Remove dev dependencies (shinyjs and emo)
  - shinyjs uses dev version only for the upload success indicator. As far as I can recall, there's no error if you have an older version, it just doesn't show the indicator. So we could remove the requirement for the development version, but keep it in the lockfile.
  - emo is not on CRAN at all, currently. Might be simple enough to re-implement the functionality we need.
  - Some jsonvalidate functionality is in the dev version
- [x] Ensure methods have all arguments of the generics (see [writing R extensions ch. 7](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Generic-functions-and-methods))
- [x] Check language in [DESCRIPTION](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#The-DESCRIPTION-file) and function docs: title case where it's needed, acronyms explained, URLs in angle brackets, etc.
- [x] Drop dev version number (x.x.x.9008)
- [x] Check on [rhub](https://builder.r-hub.io/)

karawoo:
See also https://github.com/DavisVaughan/extrachecks

karawoo:
bumping priority as I'll be talking about submitting to CRAN in the code review meeting

karawoo:
For examples, I think that `\dontrun{}` is appropriate but some sources recommend instead writing a function to check if the necessary stuff is available and putting the example in a conditional. I think let's try as-is and if there's a problem we can change it.

karawoo:
[...]

PR #253, opened 2020-01-10
Update examples in documentation

karawoo:
Part of #217. All functions in .Rd files need to have an `examples` section (even if they are not exported). And [tidyverse style](https://style.tidyverse.org/documentation.html#internal-functions) recommends not creating .Rd files for unexported functions, which we were previously doing. This PR:

- Removes .Rd files for unexported functions
- Exports a few functions that I do think ought to be documented/accessible
- Adds examples to all (?) functions that are documented in .Rd files

@Aryllen will you please double check that we are not missing examples for any functions in the .Rd files?


karawoo:
Thanks @Aryllen! Both of those functions had examples that were just in the wrong location, I've fixed them.

PR #255, opened 2020-01-11
Remove development dependencies

karawoo:
Part of #217. We currently depend on 3 packages that are not on CRAN (or are not yet updated to the version we use). This PR removes the dependencies so we only depend on packages available on CRAN.

- emo: I've reimplemented the small amount of functionality we need from emo (just inserting a few emojis in the reporting results).
- jsonvalidate: we were using jsonvalidate 1.2.2 because of support for references in schemas, however since _remote_ references are still not supported, I don't think we are actually using this feature, thus can downgrade to the CRAN version. @kelshmo does this seem right to you?
- shinyjs: the dev version has some functionality that allows the `with_busy_indicator()` functions to work. However, this is really not crucial -- the app will still function with an old version of shinyjs. I suggest we remove the requirement of >= 1.0.1.9006, but keep using the development version in our lockfile.


Aryllen:
> * shinyjs: the dev version has some functionality that allows the `with_busy_indicator()` functions to work. However, this is really not crucial -- the app will still function with an old version of shinyjs. I suggest we remove the requirement of >= 1.0.1.9006, but keep using the development version in our lockfile.

Would it be possible to document this somewhere? That way, if the functionality is missing (but also desired), someone would be able to figure out that they just need to update to the dev version of shinyjs?

karawoo:
Good idea -- will do

karawoo:
@Aryllen I just remembered that I already documented this in #253 so I am going to revert this commit.

PR #263, opened 2020-01-16
Add check_all() to do all validation checks

Aryllen:
Fixes # N/A

- Add check_all(), which does all the validation checks and sends back a list of check/pass for each of the validation checks. Parameters:
* A tibble/dataframe with file_name, metadata_type, species, assay, file_data. The file_data is a list column. There should be all four rows present, one for each metadata_type possible.
* The config list and requires configuration settings for templates and required_columns.
* Annotations dataframe.
* Synapse client object.
- Remove validation checks from app_server().

Note: @kelshmo, the PEC config will need to be updated to have templates for every specific species that is possible for biospecimen metadata. This is due to the biospecimen template being pulled from the config by the species given. If a single template can be used for multiple species, there would be config options for every species in the species list, but the template id would be repeated. Also, the species needs to match the options in the config exactly (i.e. if "mouse or other model" is a possible species, then "mouse or other model" needs to be the reference for the template, instead of just "mouse"). (See the difference between the old config and the new config in this PR for an example.)

- Added tests for new functions or for new behavior in existing functions: Yup
- If adding a new exported function, added it to the reference section of `_pkgdown.yml`: Yup
- If adding a new configuration option, documented it in `vignettes/customizing_dccvalidator.Rmd`: N/A


Aryllen:
I changed this to ready for review. I do have one concern in that my tests for `check_all()` work locally if I run them separately, but they are failing when I run them with `devtools::check()`. Haven't figured out why this is happening, yet.

kelshmo:
PR #269 should take care of the changes required for PEC. 

@Aryllen Was the move of check functions from `app_server.R` to address a performance issue?

Aryllen:
@kelshmo No, moving the checking out into it's own function was partly to break up the long bit of code in the server function, and partly to start towards allowing multiple assays. Creating a table that holds the data and sending the table onto the checking function increases flexibility. Instead of having data in a single, inflexible variable (`assay`), we could have an arbitrary number of datasets in the table and check them all (assuming we update the function to handle that when the time comes).

kelshmo:
let's pair review this tomorrow!

Aryllen:
> let's pair review this tomorrow!

Alrighty.

R/check-all.R
```diff
@@ -0,0 +1,260 @@
...
+#' )
+#' res <- check_all(data, annots, syn)
+#' }
+check_all <- function(data, annotations, syn) {
+
+  # Get indices by type
+  indiv_index <- which(data$metadata_type == "individual")
+  biosp_index <- which(data$metadata_type == "biospecimen")
+  assay_index <- which(data$metadata_type == "assay")
+  manifest_index <- which(data$metadata_type == "manifest")
```
karawoo: Can we add a defensive check to make sure that there is only one of each for now? (obviously we'd take it out later if we add support for multiple assays)

tests/testthat/test-check-all.R
```diff
@@ -0,0 +1,186 @@
...
+  res1 <- check_all(data1, annots, syn)
+  res2 <- check_all(data2, annots, syn)
+  res3 <- check_all(data3, annots, syn)
+  res4 <- check_all(data4, annots, syn)
+
+  expect_true(all(unlist(
+    purrr::map(res1, function(x) {
+      is.null(x)
+    })
+  )))
```
karawoo: `purrr::map_lgl(res1, ~ !is.null(.x))` I think would be sufficient here to replace 

```
unlist(
    purrr::map(res1, function(x) {
      is.null(x)
    })
  )
```

R/app-server.R
```diff
@@ -474,34 +244,32 @@ app_server <- function(input, output, session) {
...
+        ## Load in data to table for validation checks
+        all_data <- tibble::tibble(
+          "metadata_type" = c(
+            "manifest",
+            "individual",
+            "biospecimen",
+            "assay"
+          ),
+          "file_name" = c(
+            files$manifest$name %||% NA,
```
kelshmo: It would be helpful if there was a comment here, something to the effect of: "If file is not entered, NA will be passed in place of file name".

R/check-all.R
```diff
@@ -0,0 +1,260 @@
...
+      data$file_name[indiv_index],
+      data$file_name[biosp_index],
+      data$file_name[assay_index]
+    ),
+    success_msg = "Manifest file contains all metadata files",
+    fail_msg = "Manifest file does not contain all metadata files"
+  )
+
+  ## List results
+  res <- list(
```
karawoo: Maybe let's name the list elements so that in testing we can subset by name rather than index (seems less fragile that way?)

tests/testthat/test-check-all.R
```diff
@@ -0,0 +1,186 @@
+context("test-check-all.R")
+
+library("tibble")
+syn <- attempt_instantiate()
+attempt_login(syn)
+annots <- get_synapse_annotations(syn = syn)
+
+test_that("check_all() returns a list of check conditions", {
+  data <- tibble::tibble(
```
karawoo: Can we add `skip_if_not(logged_in())` to the top of all tests that hit synapse?

tests/testthat/test-check-all.R
```diff
@@ -0,0 +1,186 @@
+context("test-check-all.R")
+
+library("tibble")
+syn <- attempt_instantiate()
+attempt_login(syn)
+annots <- get_synapse_annotations(syn = syn)
```
karawoo: Can we use pre-constructed annotations to speed this up a little and make it simpler, something like this: 
https://github.com/Sage-Bionetworks/dccvalidator/blob/22fde0371c8daa9c2d1da8e6b80d881c1ee2a347/tests/testthat/test-check-annotation-keys.R#L7-L14

R/check-all.R
Aryllen: Added. The check I added has a long error message. I put a `# nolint`, but I remember you mentioning that something complained about our `# nolint`s. Is this going to be an issue?

R/check-all.R
Aryllen: Having a named list seems to work fine. Yay

Aryllen:
All the requested changes have been made. In order to make the `check_all()` tests work correctly with `devtools::check()`, I added a bare-bones config.yml to the testthat folder. It only has the configs needed to for the checks themselves (template synIds, complete columns list).

Ready for a second check for acceptability.

On a side note, I had moved the PR to the dev branch, but the CRAN acceptance came through faster than expected. I moved it back to the master branch.

Aryllen:
I did make one more change. In the latest commit, I changed the name of the column referring to the file name from "file_name" to "name". I had originally used "file_name" to be more explicit in what the column contained. However, while working on dccmonitor, I remembered that Synapse uses "name" as the file name. So when dccmonitor gets the file view, it has "name" as the column. While this could be changed to "file_name", this is rather arbitrary and I think could lead to more frustration about things not working. Just using "name" across both apps makes things simpler and more accessible.

I went through and changed the column name to just "name" and fixed references to it.

Aryllen:
Similarly to the previous comment, I changed the column name 'metadata_type' to 'metadataType'. I was avoiding camel case initially, but Synapse uses camel case. Again, dccmonitor gets info from Synapse and it would be simpler to use the same variable names. The metadata type is referred to with camel case within dccvalidator already. From the user standpoint, it seems to make sense to stay consistent versus arbitrarily changing the name.

R/check-all.R
```diff
@@ -0,0 +1,264 @@
+#' Run all validation checks
+#'
+#' Runs all validation checks. Proper function requires an
```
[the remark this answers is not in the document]
Aryllen: That there are requirements in order for it to function properly. Should I just start with "Requires..."?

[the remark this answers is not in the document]
karawoo: yeah I think that would be clearer

Aryllen:
All changes made.

karawoo:
@Aryllen does this work when you run the app locally? I just tried it and am seeing an error when I try to validate a file.

```
Error: Couldn't download metadata template. Make sure you are logged in to 
Synapse and that `synID` is a valid synapse ID.FALSE
```
Screen Shot 2020-02-05 at 10 26 36 AM

Aryllen:
That's interesting. It does work for me locally. 
image

karawoo:
Does it also work for you if you choose the mouse template?

Aryllen:
I just noticed that and tried it. Nope, it fails. :-/ Hmm. Let me see what's going on.

Aryllen:
The default config still had "animal" for the template name. It needed "mouse or other animal model". I updated it.

Aryllen:
Sounds good! Thank you!

PR #267, opened 2020-01-17
Remove unused packages from Imports

karawoo:
`rhub::check()` turned this up. I'm not sure why it doesn't create a note locally when I run `devtools::check()`, but regardless, these packages are no longer used in dccvalidator. DT has been completely removed, and pkgload is used in the `app.R` in this repo, but not in the dccvalidator package itself, so both should be removed from DESCRIPTION.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -29,7 +29,6 @@ Depends:
     shinyBS
 Imports:
     config,
-    DT,
     ggplot2,
     glue,
     golem,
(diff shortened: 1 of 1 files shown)


karawoo:
Yes, agreed we can remove the lockfile, however it will have to happen after we merge the PR. The current master version of dccvalidator is in the lockfile, which lists DT as a dependency. So renv won't pick up that DT is no longer needed until we update which version of dccvalidator is being used (i.e. after this PR is merged).

PR #269, opened 2020-01-21
update pec biospecimen template input

kelshmo:
Fixes # NA

- update PEC config to work with updates to `change_all()` function

- Added tests for new functions or for new behavior in existing functions **NA**
- If adding a new exported function, added it to the reference section of `_pkgdown.yml` **NA**
- If adding a new configuration option, documented it in `vignettes/customizing_dccvalidator.Rmd` **NA**

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -88,7 +88,7 @@ pec:
       human: "syn20781820"
       animal: "syn21053116"
     biospecimen_templates:
-      mouse or other animal model: "syn20768131"
+      animal model: "syn20768131"
       human: "syn20768131"
     assay_templates:
       wholeGenomeSeq: "syn20980993"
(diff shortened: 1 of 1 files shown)


config.yml
```diff
@@ -88,7 +88,8 @@ pec:
       human: "syn20781820"
       animal: "syn21053116"
     biospecimen_templates:
-      general: "syn20768131"
+      mouse or other animal model: "syn20768131"
```
Aryllen: This should be: ` animal model: syn20768131`

The PEC config has the species "human" and "animal model". These species dictate which species options appear in the app for the user and, ultimately, which species-based templates are available.

https://github.com/kelshmo/dccvalidator/blob/bcb4b11eb608439e4e9b1cceb43ae9a4be30aebd/config.yml#L97-L99

Sorry for the confusion. Please let me know if you have any questions.

PR #278, opened 2020-02-05
Coerce logical

### LINKED ISSUE #276: Add booleans to can_coerce()
We removed the enumerated values from boolean annotations. We should update `can_coerce()` to treat any capitalization of true/True/TRUE as valid.

karawoo:
Fixes #276 

- Ensures that `true`, `True`, `TRUE`, `false`, `False`, `FALSE` are all treated as coercible to logical values. While working on this, I noticed a couple small bugs with the function as it was written and fixed those as well.

### DIFF
--- a/R/check-annotation-values.R
+++ b/R/check-annotation-values.R
@@ -332,14 +332,9 @@ can_coerce <- function(values, class) {
     return(TRUE)
   }
 
-  ## Convert factors to strings
-  if (inherits(values, "factor")) {
-    values <- as.character(values)
-  }
-
   if (class == "character" &
     (inherits(values, "numeric") | inherits(values, "integer") |
-      inherits(values, "logical"))) {
+      inherits(values, "logical") | inherits(values, "factor"))) {
     return(TRUE)
   } else if (class == "numeric" & inherits(values, "integer")) {
     return(TRUE)
--- a/R/check-annotation-values.R
+++ b/R/check-annotation-values.R
@@ -326,12 +326,18 @@ check_type <- function(values, key, annotations, whitelist_values = NULL,
 #' can_coerce(2.5, "integer")
 #' }
 can_coerce <- function(values, class) {
-  if (class == "character" &
   ## If the values are already the correct class, then obviously the answer
   ## should be yes
   if (inherits(values, class)) {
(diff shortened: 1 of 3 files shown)


tests/testthat/test-check-annotation-values.R
```diff
@@ -424,3 +424,38 @@ test_that("can_coerce() returns FALSE if values aren't coercible", {
...
+  expect_true(can_coerce("foo", "character"))
+  expect_true(can_coerce(TRUE, "logical"))
+  expect_true(can_coerce(factor("foo"), "factor"))
+})
+
+test_that("factors are converted to string before checking coercibility", {
+  can_coerce(factor("TRUE"), "logical")
+  can_coerce(factor("foo"), "character")
+  can_coerce("foo", "character")
+})
```
Aryllen: These tests are missing `expect_true()`. Since the last one isn't a factor and is checked elsewhere, it doesn't quite fit in this set, either.

karawoo: Good catch, fixed, thanks!

Aryllen:
I should have double checked that the tests passed. It looks like one of the factor ones is failing.

R/check-annotation-values.R
```diff
@@ -321,12 +326,29 @@ check_type <- function(values, key, annotations, whitelist_values = NULL,
...
     return(TRUE)
+  } else if (class == "logical") {
+    if (all(values %in% c("true", "True", "TRUE", "false", "False", "FALSE"))) {
+      return(TRUE)
+    } else {
+      return(FALSE)
+    }
   } else {
     return(FALSE)
   }
```
Aryllen: Failing on `can_coerce(factor("foo"), "character")` because `class == "character"`, but "foo" is a character, not numeric, integer, or logical. Perhaps another check at the top for `inherits(values, "character")`.

karawoo: Actually I wonder if it would be better to add `inherits(values, "factor")` and remove the lines above where we convert factors to character. Not sure why I didn't think of that. Does that sound ok to you?

Aryllen: That's a good idea.

PR #283, opened 2020-02-12
Allow validating without manifest

### LINKED ISSUE #282: Make manifest file optional
By restricting the manifest file type that can be uploaded (.tsv, .txt),the app requires the inclusion of a manifest file. 

End users may want to validate the metadata files independently.

karawoo:
Fixes #282

### DIFF
--- a/R/app-server.R
+++ b/R/app-server.R
@@ -189,7 +189,8 @@ app_server <- function(input, output, session) {
             message = "Please upload some data to validate"
           ),
           need(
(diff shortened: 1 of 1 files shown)


karawoo:
Thanks @kelshmo! Right now we are specifically preventing .csv files from being uploaded. Do you think we should expand that to other file types? I don't know what else people might try to upload, and file extension is not a perfect method for determining the type. But we could expand this line in the future to disallow other file extensions (xlsx?) if needed.

PR #289, opened 2020-02-19
Fix data_summary() bug and add test

### LINKED ISSUE #288: File summary fail if column is empty
Noticed this when trying to change the dccmonitor file summary to the new dccvalidator module. If a column is all `NULL`, then there is a replacement length 0 error.

Aryllen:
Fixes #288.

- Fix the `data_summary()` bug where empty columns resulted in an error
- Add tests

- If adding a new exported function, added it to the reference section of `_pkgdown.yml`: N/A
- If adding a new configuration option, documented it in `vignettes/customizing_dccvalidator.Rmd`: N/A

### DIFF
--- a/R/mod-file-summary.R
+++ b/R/mod-file-summary.R
@@ -179,10 +179,10 @@ data_summary <- function(data) {
   for (var in data_sum$skim_variable) {
     var_col <- which(names(data) == var)
     val_summary <- summarize_values(data[[var_col]])
-    if(!is.null(val_summary) && length(val_summary) > 0) {
+    if (!is.null(val_summary) && length(val_summary) > 0) {
       data_sum$value_occurrence[data_sum$skim_variable == var] <- val_summary
     } else {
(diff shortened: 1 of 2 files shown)


karawoo:
Would it work to use `NA` instead of `NULL` in these lines within `summarize_values()` instead? Or does that cause other problems?

https://github.com/Sage-Bionetworks/dccvalidator/blob/f4201a1f20ddcbee3d77b61778f1ea8635ebf31b/R/mod-file-summary.R#L198-L200

Aryllen:
I can try that. It's not always sending back `NULL` for all of these, though. If there's more than one empty value, it sees a list and doesn't return `NULL`. I should add a check for that.

Aryllen:
I changed it so `summarize_values()` is more stable and returns `NA` if given `NA` or `NULL`. It works for a single NULL/NA or for a list of NULLs/NAs. Updated/added tests.

Issue #311, opened 2020-03-10
Include test data files


karawoo:
Based on feedback from users, it would be helpful to make available the test files we use for demos. I'm not sure which would be the most ideal way to do this: we could put them in the Synapse project where the metadata templates are stored, or bundle them with the app in `inst/extdata`. 

Should we add a button to the app that downloads the files from wherever they're stored to the user's machine? If so, where's the best place for that? Could go in the instructions, but it might be easily overlooked there. Thoughts, @Aryllen and @kelshmo?

kelshmo:
I don't think we have the best strategy implemented - I think it is easily overlooked. Currently, a PEC user could find them in [this Wiki](https://www.synapse.org/#!Synapse:syn2787333/wiki/599260), which is private to that group. 

I don't see an obvious place for a button within the app and I think this could best live within Synapse anyway. I will move the example files, and add a more obvious link, to [where the templates are stored](https://www.synapse.org/#!Synapse:syn20729790).

cw-dvr8:
@kelshmo How about waiting until after the demo tomorrow to move the files?  I've updated all of the ones that I'm planning on using.

Aryllen:
I'm a bit torn on this. I think a "Download Test Templates" button might be able to go next to the "Show Instructions" button, but it might also confuse people. Alternatively, it could be in the "Using the App" tab under a section called something like "Try Out the App". It wouldn't necessarily need to be a button, but a link to a Synapse folder that holds the test templates.

karawoo:
Just an update that we have test files for AMP-AD here: https://www.synapse.org/#!Synapse:syn21781041

Aryllen:
One challenge with demo files is that they become out of date as we update templates. It would be handy if the demos would always be up to date with the latest version of the templates, even if that means new key columns have no values.

PR #314, opened 2020-03-11
Add more documentation of how to use dccvalidator for other projects

### LINKED ISSUE #307: Vignette on using package components
Would be good to have a vignette that describes a) how to use the checking functions directly (i.e. from within an R script), and possibly b) how to use some of the modules without using the entire app. The idea here is to provide some guidance for projects that might find some components of dccvalidator useful but don't have the same project structure that we have in AMP-AD/PEC (so the types of files we enforce for those consortia won't be as relevant).

karawoo:
Fixes #307 

Adds a lot of content to the "Customizing dccvalidator" vignette about how to reuse components of dccvalidator even if the exact structure of your project is different from what we enforce in AMP-AD/PEC. We now cover how to reuse dccvalidator's functions in scripts, how to create your own functions, and how to use the `results_boxes` module in other Shiny apps. In working on this, I noticed a place where the pkgdown documentation could be better organized to show exactly which functions are used for checking data, so I made that change as well.

For demonstration purposes I've included my 4 test files in `inst/extdata` (see also #311). I use some of these in the examples in the vignette, but we could still also host them on Synapse if we want an easier location for non R users to find them.

### DIFF
--- a/vignettes/customizing-dccvalidator.Rmd
+++ b/vignettes/customizing-dccvalidator.Rmd
@@ -19,7 +19,10 @@ built-in Shiny application is designed to ensure that data conforms to a
 specific project organization structure in which data is organized into studies
 that are described with a combination of metadata files. These metadata files
 document the individuals, specimens, and assay(s) that were performed in the
-study.
+study. Documenting the data in this way allows researchers to describe the study
+at different levels while minimizing repetition. These metadata files can later
+be joined on the `individualID` and `specimenID` columns to create a table of
+all of the metadata for the study.
 
 For projects like AMP-AD, PsychENCODE, and others that follow the same
 structure, see the next section ("Customizing the Shiny application") for how
@@ -109,7 +112,8 @@ To install the dccvalidator instead of forking the repository:
   biospecimen template the app validates against.
 * `complete_columns`: For each metadata file and the manifest, a list of the
   columns that must be complete (i.e. not contain any missing values or empty
-  strings)
+  strings). For metadata files, this should typically include `"individualID"`
+  and/or `"specimenID"`.
 * `contact_email`: Email address linked in footer for users to contact if they
   have questions
 
--- a/vignettes/customizing-dccvalidator.Rmd
+++ b/vignettes/customizing-dccvalidator.Rmd
@@ -14,8 +14,24 @@ knitr::opts_chunk$set(
 )
 ```
 
-dccvalidator is intended to be customizable for different settings. The Shiny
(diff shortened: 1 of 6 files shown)


vignettes/customizing-dccvalidator.Rmd
```diff
@@ -14,8 +14,24 @@ knitr::opts_chunk$set(
 )
 ```
 
-dccvalidator is intended to be customizable for different settings. The Shiny
-application uses a [configuration file](https://github.com/Sage-Bionetworks/dccvalidator/blob/master/config.yml)
+dccvalidator is intended customizable for different settings, however the
+built-in Shiny application is designed to ensure that data conforms to a
+specific project organization structure in which data is organized into studies
+that are described with a combination of metadata files. These metadata files
+document the individuals, specimens, and assay(s) that were performed in the
```
kelshmo: It could be useful to add some commentary describing the design thinking for this metadata structure - to join across files on individualID and specimenID variables. Might be useful to note those variables are defined in the `complete_columns` configuration option.

vignettes/customizing-dccvalidator.Rmd
```diff
@@ -97,3 +113,168 @@ To install the dccvalidator instead of forking the repository:
...
+See the [function reference](https://sage-bionetworks.github.io/dccvalidator/reference/index.html#section-data-validation-functions) for a complete list.
+
+These functions are used in the dccvalidator Shiny app, but they can also be
+used in scripts or reports. Each function takes data as input and returns
+results in the form of custom condition objects. The condition objects inherit
+from R's `"message"`, `"warning"`, and `"error"` classes. However, rather than
+raising a message/warning/error, the check functions in dccvalidator return the
+condition objects themselves.
+
+```{r demo-check-function, message = FALSE}
```
kelshmo: this is great!

vignettes/customizing-dccvalidator.Rmd
karawoo: Good idea! I've added more context, but I didn't want to mention the config file until the "Customizing dccvalidator" section. So I have instead added more information to the documentation of the `complete_columns` option further down in the document.

karawoo:
> would the action be to modify the file input to remove a file type from the validator tab or would the modification be more complicated?

You could remove a file type from the validator fairly easily, and we already do this: if you select "drosophila" as the species in AMP-AD, the individual metadata input box disappears. In practice, I'm guessing people might want to make other changes (like removing the species and study inputs), and that would be slightly more complicated.

PR #317, opened 2020-03-17
Add config for annotations user research

karawoo:
Adds a config that is identical to AMP-AD except saves to a different project and allows anyone on either the "Sage Bionetworks" or "Annotations Design Research Participants" teams to upload.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -131,3 +131,38 @@ pec:
       - "libraryPreparationMethod"
       - "runType"
   contact_email: "kelsey@synapse.org"
+
+annotations-research:
+  parent: "syn21782020"
+  annotations_table:
+    - "syn10242922"
+    - "syn21459391"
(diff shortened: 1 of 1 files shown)


config.yml
```diff
@@ -131,3 +131,38 @@ pec:
       - "libraryPreparationMethod"
       - "runType"
   contact_email: "kelsey@synapse.org"
+
+annotations-research:
+  parent: "syn21782020"
```
Aryllen: I don't have access to this folder. May want to set permissions for the same teams.

karawoo: It is accessible to the Sage Bionetworks team, can you request to join? You should be on there anyway. https://www.synapse.org/#!Team:273957

Aryllen: Oh, I thought I was on that team. I just requested access. Thanks!

PR #320, opened 2020-03-20
Remove AMP-AD team from default settings

Aryllen:
Fixes # NA

AMP-AD doesn't have permissions to create folders in the validation-sandbox project, which causes the app to crash if they log in while the config is set to default. Ultimately, it would be better to have a default view that shows it's not the AMP-AD version. This is a quick solution, but will not result in much less confusion since it will still tell the user that they need to be a part of the AMP-AD team.

@kelshmo What do you think? Should we push this change and make a separate issue to update the default config items (including a default vignette) when we get a chance? Or should we prioritize getting the default items done on this branch and push this change as a whole?

- Remove AMP-AD team from the default teams in config.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -9,7 +9,6 @@ default:
   study_link_human: "https://adknowledgeportal.synapse.org/#/Explore/Studies?Study=syn3159438"
   study_link_ref: ""
   teams:
-    - "3320424"
     - "273957"
   templates:
     manifest_template: "syn20820080"


kelshmo:
Since they don't have permissions to create folders in the validation-sandbox, I agree the team should be removed from the default config. 

It is not clear to me what else you are proposing here - I think a default vignette would be useful, but lower priority.

Aryllen:
> It is not clear to me what else you are proposing here - I think a default vignette would be useful, but lower priority.

As it is now, if the app is set to the default configuration, it appears as if the app is the AMP-AD version (but it's not). So if the configuration somehow gets reset in the future, AMP-AD contributors would see the expected AMP-AD view, but not be able to use the app. They would also get a message saying that they needed to be a part of the AMP-AD team (which they already are). With a default app appearance, it would be simple to debug the issue: notice that the app looks like the default version and then update the config file to be AMP-AD again. 

We could also use the default vignette for "advertising" how to customize the application for other consortia.

I do agree that it's probably lower priority than simply removing the group from the config teams. I'll push this change. Thanks!

kelshmo:
I'm with you - yes, we should also remove the configuration options that assign the AMPAD vignette to the default app. I agree that is confusing!

PR #339, opened 2020-04-07
Updates to AMP-AD documentation, including adding general process overview

### LINKED ISSUE #324: Improve AMP-AD instructions
PEC has a fairly thorough set of instructions, and it would be nice to do the same for AMP-AD. There seems to be a fair amount of confusion as to the exact steps the contributors are to take, including not informing the DCC for verification of proper validation before trying to upload. They are also not uploading with the manifest.

Aryllen:
Fixes #324 -- sort of. It's hard to tell how detailed the AMP-AD **Using the App** section needs to be because it's unknown if it's being read in the first place. I do think that the onboarding videos that are currently in discussion may help with keeping this document simple. More specifically, a video on metadata files and the manifest file would reduce the need to expand on these in a future update to this document.

- Change what was **Instructions** to **Requirements**. Add new **Instructions** header after this section.
- Add a **General Process Overview** section. I know this is a bit broad, but if we are referring people to the dccvalidator for the instructions/information, then I think we should be transparent in what the overall process is. I don't think this has to be detailed for each individual case, but just an overall idea of what should happen and when.
- Add file format information to the **Metadata Requirements** section.
- Fix text in **Viewing Data Summary** section to not be centered.
- Title case the headers.

@karawoo, @amapeters, any thoughts/improvement ideas on this?


inst/using-the-dccvalidator-app-amp-ad.Rmd
```diff
@@ -38,7 +38,22 @@ Some portions of the app submit data to Synapse. This allows curators at Sage to
 troubleshoot issues if needed; no one outside the Sage curation team will be
 able to download the data.
 
-## Documentation upload
+# Instructions
+
+## General Process Overview
```
karawoo, line 43: I might suggest we mention explicitly that this section has the overall overview of the process and that the following sections cover each of these steps in more detail.

Aryllen: I've added a blurb here. I'm not sure if it's not clear enough. If it's not, I could change it to a list, like:

> The following sections are covered:
> - **General Process Overview** an overview of the data contribution process.
> - **Documentation Upload** how to upload study and assay documentation.
> - **Data Validation: Metadata Requirements** requirements for the metadata and manifest files and where to find templates.
> ...

Not sure which would be better, a blurb or a table of contents style.

Aryllen:
Great! Thank you for the feedback!

PR #346, opened 2020-04-17
Check for duplicate file paths in manifest

### LINKED ISSUE #343: Add check for no duplicate paths in manifest
Had a contributor who did not understand the manifest file and believed there should be one row per individual/specimen for each metadata file. Their files passed because it was filled out, but it was also incorrect.

All paths should be unique in the manifest. Adding this check would have at least given a notice to the contributor that something was wrong.

Granted, the file paths were also not real paths to files, but I'm not sure how to fix that problem, yet (one idea would be run syncToSynapse as a dryRun and see if that succeeds, but there's potential issues with that).

karawoo:
Fixes #343 

Adds check for duplicate file paths. This only checks if file paths are written identically -- it will not catch the same path written multiple ways (e.g. `"~/file.txt"` vs. `"/Users/me/file.txt"`). We don't have access to the user's filesystem, so I can't think of a way around this.

This also does not check whether the file _names_ are unique within a given Synapse parent. For example, the following manifest would be considered valid according to this check, but on upload this file would be uploaded as two versions of the same file, rather than as two separate files:

| path| parent |
| ------------- | ------------- |
| /folderA/file.txt | syn123 |
| /folderB/file.txt | syn123  |

It might be worth checking for this case too, but maybe as a separate check.

- If adding a new exported function, added it to the reference section of `_pkgdown.yml`: yes


R/check-duplicate-files.R
```diff
@@ -0,0 +1,52 @@
...
+        behavior = behavior,
+        data = NULL
+      )
+    )
+  }
+  if (anyDuplicated(stats::na.omit(data$path))) {
+    check_fail(
+      msg = fail_msg,
+      behavior = behavior,
+      data = data$path[duplicated(data$path)]
```
karawoo, line 44: Need to remove NAs from here as well

R/check-duplicate-files.R
```diff
@@ -0,0 +1,52 @@
...
+#' @return A condition object indicating whether the data contains duplicated
+#'   file paths in the `parent` column.
+#' @export
+#' @examples
+#' manifest <- data.frame(
+#'   path = c("/path/to/file.txt", "/path/to/file.txt"),
+#'   parent = c("syn123", "syn123")
+#' )
+#' check_duplicate_files(manifest)
+check_duplicate_files <- function(data,
```
karawoo, line 24: Change to `check_duplicate_paths`

karawoo:
Yup I added a test for that too

karawoo:
Yeah, I think that makes sense. I've updated it so it'll warn if the path column is all NAs.

Issue #352, opened 2020-04-28
check_ages_over_90 error when all NA


Aryllen:
`check_ages_over_90()` is breaking when the values in `ageDeath` are all `NA`. The error is that there is a missing value where true/false is needed.

It's probably this line: https://github.com/Sage-Bionetworks/dccvalidator/blob/dc3d97a94b9f7056b726fc69525380110f3531e3/R/check-ages-over-90.R#L39

karawoo:
I should have added a test for that, sorry, I can fix

Aryllen:
No worries, and thank you!

karawoo:
reprex: 
``` r
library("dccvalidator")
#> Loading required package: shinyBS
dat <- data.frame(ageDeath = c(NA, NA))
check_ages_over_90(dat)
#> Error in if (!any(purrr::map_lgl(age_data, ~any(is_over_90(.x))))) {: missing value where TRUE/FALSE needed
```

<sup>Created on 2020-04-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

PR #354, opened 2020-04-28
Don't return NAs in final data

karawoo:
#353 fixed the bug in #352 but returned NAs in the final data.

``` r
dat <- data.frame(ageDeath = c(NA, 95))
res <- check_ages_over_90(dat)
res$data
#> $ageDeath
#> [1] NA 95
```

<sup>Created on 2020-04-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

This fixes that by updating `is_over_90()` so it considers `NA` not greater than 90 rather than returning NAs.

### DIFF
--- a/R/check-ages-over-90.R
+++ b/R/check-ages-over-90.R
@@ -36,7 +36,7 @@ check_ages_over_90 <- function(data, col = "ageDeath", strict = FALSE,
   cols_in_data <- intersect(col, names(data))
   age_data <- data[, cols_in_data, drop = FALSE]
 
-  if (!any(purrr::map_lgl(age_data, ~ isTRUE(any(is_over_90(.x)))))) {
+  if (!any(purrr::map_lgl(age_data, ~ any(is_over_90(.x))))) {
     check_pass(
       msg = success_msg,
       behavior = behavior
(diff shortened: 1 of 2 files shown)


R/check-ages-over-90.R
```diff
@@ -55,8 +55,8 @@ check_ages_over_90 <- function(data, col = "ageDeath", strict = FALSE,
 }
 
 # Does the column (after removing non-numeric characters) contain any values
-# >90?
+# >90? NAs are not considered over 90 and will evaluate to FALSE.
 is_over_90 <- function(x) {
   col_numeric <- gsub("[^0-9]", "", x)
-  as.numeric(as.character(col_numeric)) > 90
+  (as.numeric(as.character(col_numeric)) > 90) %in% TRUE
```
Aryllen, line 61: This broke my brain for a second. XD

karawoo: This is my solution for creating a vectorized version of `isTRUE()`

Aryllen: It makes sense. Just took me a second to get it, haha.

Issue #257, opened 2020-01-13
Warn about duplicated individual or specimen IDs across studies


karawoo:
Once we have a master table of individuals and specimens, add a warning if the uploaded data reuses individuals or specimens that are already in use elsewhere in the consortium. The contributor should verify that if they are reusing IDs it is because they actually refer to the same individuals and specimens.

PR #382, opened 2020-05-22
Validate that additions to existing studies retain all IDs that were previously shared

### LINKED ISSUE #55: Support validating additions to existing datasets
Sometimes we get new data/metadata for an existing study. The validation app doesn't know about the existing data, so it might report erroneous mismatches between the metadata and manifest (for example, if the metadata contains specimens that are not in the manifest, but correspond to existing data). Need to think through how to support this. One possibility is sketched out below:

IMG_9783

karawoo:
Fixes #55

- Adds `check_complete_ids()` function for checking data against a table of all the samples/individuals in the consortium (I chose to name it `check_complete_ids()` instead of `check_existing_ids()` because the latter seems more vague)
- Updates `check_all()` to run this check on the individual, biospecimen, and assay metadata. `check_all()` gains new arguments in order to only run the check for existing studies
- The `get_study_names()` module now returns a list of the study name and whether or not the study already exists (previously it returned just the name). This update allows us to pass the `study_exists` parameter to `check_all()`.

One thing that is not super ideal is that if a study exists, but is not in the master table, then we see a repeated warning for each file:

Screen Shot 2020-05-22 at 3 16 53 PM

I'm not sure the best way to avoid this without adding arguments to `check_complete_ids()` for every possible message. Maybe before reporting the results we should call `unique()` on the list returned by `check_all()`? What do you think @Aryllen?

@kelshmo, do you need this check? Or does PEC only need #257? If this shouldn't be enabled for PEC, then we need to tackle the issue of how to run different checks in different consortia.

- If adding a new exported function, added it to the reference section of `_pkgdown.yml`: yes


Aryllen:
> One thing that is not super ideal is that if a study exists, but is not in the master table, then we see a repeated warning for each file

This is something that has been in the back of my mind recently. The question, "does the study currently exist?" is not quite answering the question that we need answered for this problem. I think it may also add confusion, especially when we go to update the validator to have a flow chart showing what should be done if a study currently exists or if a study is new.

A study appearing in that list only means that the study name has been added to the study table. This does not mean that there is existing data or information for that study (i.e. the real question we need answered). We also have the issue that some people are entering some other name for a study that does exist, sometimes because I have asked them to (we have two people adding to MSBB so I have requested that they enter different names in the validator). These study names will not be captured in the master table, even though they do exist.

I am trying to think of a way to fix this. Ultimately, I think the study name question needs to be more clear, but the rest is a challenge. Here are some thoughts:
- We could make `study$study_exists` based on a check if the study name is in the master table, instead of being the answer to the question. That does not help the cases where it does exist, but the study name given is different.
- We could also have a second question for whether they are uploading additional data for a study. If yes, a conditional panel pops up that requests the name of the study, which is then checked for in the master table. This might be confusing for people who are uploading publication data that is analyzing data from a different study (or studies), though.

kelshmo:
This worked beautifully for me when a study exists in the master table, I have a [dev version of the app here](https://shinypro.synapse.org/users/kmontgomery/dev/dccvalidator-dev/). We do need this check in PEC since we have additions to existing studies and the workflow is the contributor takes the existing individual/biospecimen/assay metadata file, downloads it, alters it and re-uploads it. That said, we do not have a master table currently, so I would not update our fork until I have taken care of that. 

When a study does not exist in the master table, does it make sense to surface that warning "Can't check for complete IDs because master table does not contain..."? Maybe the study exists but no data has been added to it yet? I'm not sure the user will know what action to take with that message.

karawoo:
> We could make study$study_exists based on a check if the study name is in the master table, instead of being the answer to the question. That does not help the cases where it does exist, but the study name given is different.

I like this idea, and then we can do away with the warning about the study not being found in the mater table. @kelshmo is right that it's not really actionable for the contributor (it may be actionable for us if the study should be there but is missing, but maybe that's a separate issue).

This still leaves the case of what to do if they're using a study name that's different from the one they will ultimately upload to. What is the reason for this @Aryllen? Is it to get around #197? If so, how are you going to reconcile these different files? Maybe we need to raise the priority of that issue and find permanent solution, rather than work around it in this one.

Aryllen:
> What is the reason for this @Aryllen? Is it to get around #197?

Yes, in that it is somewhat related to this issue of metadata management, and no, in that it's both a user problem and a dccmonitor problem. We still need a way to solve #197. The tl;dr version is that this might be two separate issues, but in the meantime will affect the results from this check.

For the dccmonitor problem, for this release I have a case where two people are uploading data for the same study. I requested that they validate under different study names so that dccmonitor can distinguish between their sets of files. If they validated under the same study name, dccmonitor would only show me the most recent version of whichever metadata file was uploaded. You are correct in that we should solve the problem in dccmonitor versus find a work around here, however. Now that the validation is a button click, I could look into adding the ability to select which files that were uploaded for validation ([Issue 52](https://github.com/Sage-Bionetworks/dccmonitor/issues/52)).

The user problem is different in that I have users that enter in different study names. I have at least three users I can think of off the top of my head that used at least 2 different study names for validation of study data, whether it was typing in two different study names (one I think added an extra space which could be addressed in an update), or a mix of choosing the study from the drop-down but later validating via typing in a name. I have also had a user not choose their existing study from the drop-down, but rather just made up their own study name. In the case of user error, they would not get the validation checks against the master table. 

I honestly do not know how to solve the user problem or even what the problem is. I am only guessing that the question itself is confusing them (although I think it will be confusing when we give a flow chart for new versus existing studies). Fixing the question to be more understandable or both fixing it and adding a second question might be a separate issue.

karawoo:
I wonder if we could lessen the cognitive load on users by removing the "is this an addition to an existing study?" question altogether and having just a dropdown listing the studies, with an "Other" option that would then trigger a write-in box? Though with the number of studies we have, maybe the list would make it too hard to find the "Other" option

Aryllen:
> maybe the list would make it too hard to find the "Other" option

I had this same thought. The list is in alphabetical order so they could logically find it in the 'O' section, or we could add it to the end of the list. But either way, I am not sure they will find it in every case. It would also make it more annoying for us, I think, since testing files is pretty fast without having to search for 'other' before being able to type in a test name (although ease for us is less important).

Hmm...I am not quite sure what the best option would be. I like this idea. Just trying to think if it might add yet another problem. Should we open a separate issue for updating the question section of the app (with whatever we decide on)?

karawoo:
Yeah, let's open another issue. Maybe we could have it be a free-text box that as they type will match to existing study names that they could then select, or if there's no match then it just gets saved as whatever they typed?

Aryllen:
> Maybe we could have it be a free-text box that as they type will match to existing study names that they could then select, or if there's no match then it just gets saved as whatever they typed?

Oh, that's a neat idea. Might have more typo issues, but might not. Anywho, I will open an issue for this topic.

Will you be updating the study existence to be dependent on being in the master table instead of the answer to the question?

karawoo:
Yes I will

Aryllen:
Great! Thank you. I think the rest seems fine. I will check it over again after the update.

karawoo:
@Aryllen this is ready to be re-reviewed. `check_all()` now determines whether to run `check_complete_ids()` based on the contents of the sample table. `check_complete_ids()` errors instead of returning `"check_warn"` if the samples table it receives doesn't contain the study. In practice this won't happen in the app, because the function is only run if the table has the study.

Aryllen:
I think the next conversation is when to merge this. @kelshmo doesn't have a specimen/individual ID table, yet. We could merge and PEC could not update their dccvalidator, but if we have more new updates to the validator before they get a table, then it would be a shame not not be able to update to those.

kelshmo:
I will prioritize dev of the table in July. You can merge now.

PR #390, opened 2020-06-18
Update for CRAN release

### LINKED ISSUE #387: New CRAN Release Required
See message below. It looks like shinyjs and jsonvalidate both use V8 (although jsonvalidate is only a suggests for us). Since neither are on the list, I assume they have updates that remove this dependency. Should update the packages to the latest versions (and require these versions).

> Package V8 no longer installs on fedora-clang after upgrading to Fedora 32.
> 
> Packages
> 
> codemetar datarobot dccvalidator dm dqshiny ggforce mcmcabn shinyMobile
> sparrpowR textplot
> 
> suggest it or a package which requires it but do not use it
> conditionally (see §1.1.3.1 of 'Writing R Extensions').
> 
> Please correct ASAP and before June 30 to safely retain the package on CRAN.

Aryllen:
Fixes #387 (hopefully)

- Add conditional for running the tests for the json validation check. This should get around jsonvalidate not being available in Fedora-clang (on Fedora 32) due to the V8 package (thanks, @karawoo, for helping me with that).
- Add NEWS.md
- Update PR template to have check boxes to reduce the number of NAs that have to be typed (lol). Most updates don't need all of the below items.
- Update PR template to request updating the NEWS.md file if functionality is added or altered. Would not need to update it for things like updating the config (i.e. adding new templates, required columns, etc), unless it was adding or removing a config option.
- Update to v0.3.0

Still need to update cran-comments.md, but currently running OS checks.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,3 @@
-# dccvalidator (development version)
-
-- (develop branch) Add `check_complete_ids()`, which includes new config option designating the table to compare against for existing specimen and individual IDs
-
 # dccvalidator v0.3.0
 
 - Add `check_parent_syn()`
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,0 +1,21 @@
+# dccvalidator (development version)
+
+- (develop branch) Add `check_complete_ids()`, which includes new config option designating the table to compare against for existing specimen and individual IDs
+
+# dccvalidator v0.3.0
+
+- Add `check_parent_syn()`
+- Add modal message if `sessionToken` is invalid
+- Add `check_ages_over_90()`
+- Add `check_duplicate_paths()`
+- Update `can_coerce()` to be `TRUE` for any data type needing to be coerced to `character`
+
+# dccvalidator v0.2.0
+
+- Move all validation checks to `check_all()`
+- Move data file summary to module
+- Update `can_coerce()` to include numeric to integer and character to logical
+
+# dccvalidator v0.1.0
+
+Initial release of dccvalidator package.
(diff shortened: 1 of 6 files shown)


tests/testthat/test-check-schema-df.R
```diff
@@ -17,6 +18,7 @@ test_that("check_schema_df passes if all rows valid", {
 })
 
 test_that("check_schema_df fails if all rows invalid", {
+  skip_if_not(requireNamespace("jsonvalidate", quietly = TRUE))
```
karawoo, line 21: I realized after we talked that there's already a `skip_if_not_installed()` function, which would be shorter, but this works too

NEWS.md
```diff
@@ -0,0 +1,21 @@
+# dccvalidator (development version)
+
+- (develop branch) Add `check_complete_ids()`, which includes new config option designating the table to compare against for existing specimen and individual IDs
```
karawoo, line 3: I would recommend taking this section out, rebasing develop onto master (so the news file gets added to develop) and then adding this news bullet only on develop.

Aryllen: That is a good idea for just having that news bullet on develop.

That whole news section for the development version was intended for incremental additions on master between the latest release and the present moment (idea taken from Jim Hester's lintr package NEWS.md). The PR template update included in this PR has a request to add info there if there is a change to functionality. I would still like to keep that ability so we can track the most relevant changes in a more visible way. Would it be better to remove that whole section for this specific PR, do all the CRAN/release stuff, then do another PR just adding that top header again for future updating?

karawoo: Yes, I think that would be ideal 👍

tests/testthat/test-check-schema-df.R
Aryllen: Haha, I should have noticed that when I was doing it, too, especially with auto-complete. I like `skip_if_not_installed()` better since it is more clear. I might change to that. It shouldn't negatively affect the OS tests.

PR #401, opened 2020-08-21
Fix infobox bug

### LINKED ISSUE #368: Ignore NA in info box counts
We should ignore NA when populating the information box counts of specimens/individuals. While it would still signal something was wrong if this number was something unexpected due to NAs, since we get the unique values, this would still not give an indication of how many NAs. Instead it would just have one extra regardless of 1 NA or 100 NA.

Labeling as a bug since I don't think this is the desired behavior, but it's not technically a bug in the sense of being a breaking issue.

Aryllen:
Fixes #368

- Count individualID and specimenID differently. Instead of getting the number of strings returned from `unique()`, get a vector of `TRUE`/`FALSE` by asking if those strings are `!is.na()` and sum.

- [ ] ~If adding a new exported function, added it to the reference section of `_pkgdown.yml`~
- [ ] ~If adding a new configuration option, documented it in `vignettes/customizing_dccvalidator.Rmd`~

### DIFF
--- a/R/app-server.R
+++ b/R/app-server.R
@@ -301,13 +301,11 @@ app_server <- function(input, output, session) {
     ## Counts of individuals, specimens, and files
     output$nindividuals <- renderValueBox({
       valueBox(
-        sum(!is.na(unique(
-          c(
-            indiv()$individualID,
-            biosp()$individualID,
-            manifest()$individualID
-          )
-        ))),
+        count_unique_values(
+          indiv()$individualID,
+          biosp()$individualID,
+          manifest()$individiualID
+        ),
         "Individuals",
         icon = icon("users")
       )
@@ -315,9 +313,11 @@ app_server <- function(input, output, session) {
 
     output$nspecimens <- renderValueBox({
       valueBox(
-        sum(!is.na(unique(
-          c(biosp()$specimenID, assay()$specimenID, manifest()$specimenID)
-        ))),
+        count_unique_values(
+          biosp()$specimenID,
+          assay()$specimenID,
+          manifest()$specimenID
+        ),
         "Specimens",
         icon = icon("vial")
       )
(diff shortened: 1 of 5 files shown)


R/app-server.R
```diff
@@ -301,39 +301,33 @@ app_server <- function(input, output, session) {
...
     output$nindividuals <- renderValueBox({
       valueBox(
-        length(
-          unique(
-            c(
-              indiv()$individualID,
-              biosp()$individualID,
-              manifest()$individualID
-            )
+        sum(!is.na(unique(
```
karawoo, line 304: Since we repeat this calculation in several places, I wonder if it would make sense to write a small function, something like this: 

```r
count_unique_ids <- function(...) {
  sum(!is.na(unique(c(...))))
}
```

Aryllen: I had that same thought, but haven't tried it out just yet. I vaguely remember some issues popped up when I tried doing something like that in the dccmonitor app. I'll give it a whirl and see how it goes. I would much rather have a small, separate, testable function.

Aryllen:
Added un-exported helper function `count_unique_values` in utils and a test for it. The counts boxes appear to be working nicely.  Using it for counting the IDs and the number of files in the manifest so made it slightly more generic named.

R/app-server.R
```diff
@@ -301,14 +301,10 @@ app_server <- function(input, output, session) {
...
-            c(
-              indiv()$individualID,
-              biosp()$individualID,
-              manifest()$individualID
-            )
-          )
+        count_unique_values(
+          indiv()$individualID,
+          biosp()$individualID,
+          manifest()$individiualID
```
karawoo, line 307:
```suggestion
          manifest()$individualID
```

Aryllen:
Nice catch, @karawoo. 🤦

PR #406, opened 2020-08-31
Add reset button

### LINKED ISSUE #379: Add reset button
Reset button should clear all files from the input boxes. Should it clear the study name (which might be kind of annoying)?

Aryllen:
Fixes #379

This is definitely more challenging than I expected, especially since we have modules. The current version of this branch (
[0a853aa](https://github.com/Sage-Bionetworks/dccvalidator/pull/406/commits/0a853aa441f259f20dabd757d19e10579fd00ea4)) has a button on the validation page that will reset the file inputs and success boxes.

Challenges/questions:
1. The button indicator is not fully modular and this rears its ugly head here. The indicator would be tied to both buttons, but only appear in one location (dependent on the button it's called on, which is harder to explain in text). This is why I have the button under a horizontal line. This leaves some distance for error messages from validation. There is no indicator on the reset button. Fortunately, there is still some visual feedback that shows something happened (i.e. file inputs reset). Does this layout look fine? Additionally, is it sufficient for the elements to reset (instead of having an indicator check-mark to indicate it did something)?
2. The study name module needs a way to reset. I could do this by calling it again, but it would be less efficient because it will regenerate the list of study names. Ideally, we should be getting the list of study names outside of the module and pass them in. Then I could use that variable again to reset without getting all of that information, again. However, do we want the study name to also reset or would this be more annoying?
3.  How important is it that the radio buttons for species and the assay drop down go back to defaults? It's not impossible, but it seems unnecessary. Would this end up being more annoying?

- Add reset button to validation section that resets: 
  - [x] file inputs
  - [x] success boxes
  - [x] study name module
  - [x] species
  - [x] assay
- Add reset button to documentation section that resets:
  - [x] file inputs
  - [x] study name module

~~- [ ] Added tests for new functions or for new behavior in existing functions~~
~~- [ ] If adding a new exported function, added it to the reference section of `_pkgdown.yml`~~
~~- [ ] If adding a new configuration option, documented it in `vignettes/customizing_dccvalidator.Rmd`~~

### DIFF
--- a/R/app-server.R
+++ b/R/app-server.R
@@ -325,6 +325,9 @@ app_server <- function(input, output, session) {
         )
 
         callModule(results_boxes_server, "Validation Results", res)
+
+        # Give next step if no failures
+        next_step_modal(res, config::get("contact_email"))
       })
     })
 
@@ -390,4 +393,31 @@ app_server <- function(input, output, session) {
       )
     })
   })
+}
+
+#' Modal with next step
+#'
+#' If none of the checks inherit `check_fail`, then pop up a modal to tell
+#' users what to do next.
+#'
+#' @param results List of conditions
+#' @param email Contact email as a string
+#' @noRd
+next_step_modal <- function(results, email) {
+  is_failure <- purrr::map_lgl(results, function(x) {
+    inherits(x, "check_fail")
+  })
+  if (!any(is_failure)) {
+    showModal(
+      modalDialog(
+        title = "Great work!",
+        HTML(
+          glue::glue(
+            "Your validated file(s) had no failures. Please contact <a target=\"_blank\" href=\"{email}\">{email}</a> to proceed with the next step if you have validated all finalized metadata and manifest files at once. For multiple assays, please validate each assay with your other metadata files (individual and/or biospecimen) and manifest." # nolint
+          )
+        ),
+        easyClose = TRUE
+      )
+    )
+  }
 }
--- a/R/app-server.R
+++ b/R/app-server.R
@@ -71,11 +71,11 @@ app_server <- function(input, output, session) {
         )
       )
 
+      all_studies <- get_study_names(reactive(config::get("study_table")), syn)
       study_name <- callModule(
         get_study_server,
         "study",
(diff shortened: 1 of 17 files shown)


Aryllen:
> My sense is option 2 solves the main issue and is easier. It does, however, open the possibility of someone forgetting to change the study name and validating files under the wrong name. I'm not sure how concerned we should be about that...

This is a good point because having the correct study name is really important for the dccmonitor to work correctly. The MODEL-AD team, for example, seems to have people that are working on multiple studies. If they forget to change the name, this would cause issues with us validating their results, etc.

I'd say that issue might mean leaning towards resetting all the elements...

Aryllen:
Added a helper function for resetting any number of shiny elements by their ID names. This function is not tested (nor was documentation generated, yet) mostly because I am at a loss of how to since it's main functionality is to call `shinyjs::reset()`.

Documentation tab has a reset button now. It resets the file inputs.

Still working on reset for the study names, species, and assay.

Aryllen:
I think this is finally ready for review, @karawoo.

For the latest two PRs, I'll update the dev version on each before merging so there isn't a conflict.

R/app-server.R
```diff
@@ -94,16 +95,45 @@ app_server <- function(input, output, session) {
...
         parent_folder = reactive(created_folder),
-        study_table_id = reactive(config::get("study_table")),
+        study_names = all_studies,
         synapseclient = synapse,
         syn = syn
       )
     }
 
+    ## Reset fileInputs if reset button pressed
+    ## Need to somehow reset the study name?
```
karawoo, line 105:
```suggestion
    ## Reset fileInputs, study name, and other inputs
```

Suggest tweaking the wording of this since the study name reset is working

karawoo (approved):
Nice!! This is working great. I made a tiny suggestion on the wording of a comment that seemed outdated, but this seems fantastic.

Aryllen:
Thanks, @karawoo!

I totally forgot to pull the committed suggestion and ended up with an extra merge commit. 🤦 I got too excited. 😂

Issue #377, opened 2020-05-20
Add how to use information in data summary section


Aryllen:
Add some commentary on how to use the data summary section.

Based on feedback from Duminda.

Aryllen:
Consider putting summary below validation.

PR #407, opened 2020-09-01
Add info to the results boxes explaining how to interpret them

### LINKED ISSUE #375: Give more info about check outcomes
Somehow document the difference between the different check outcomes. Perhaps with more tool tip-style pop-ups that are hidden until they are wanted, or a snippet of text near the top of the box. The idea is that some people are not necessarily sure what to do about warnings, to be specific. Need a way to let them know that these are not critical. In many cases, they can be ignored, but we don't want to just tell them to ignore these altogether. Need a way to describe what to do that is helpful to the user and not detrimental to the curators.

karawoo:
Fixes #375

Each results box now has a footer that explains how to interpret the box. These appear on startup (before results are populated) and persist once the check results have been generated. One thing that might be weird is that the success box message says "Your files passed these checks. Well done!" before there are any successes. Should I change the wording? The box is collapsed by default, so maybe it doesn't really matter.

- [ ] Added tests for new functions or for new behavior in existing functions: we don't have any functions for this module yet and I didn't add them, but let me know if you want them
~- [ ] If adding a new exported function, added it to the reference section of `_pkgdown.yml`~
~- [ ] If adding a new configuration option, documented it in `vignettes/customizing_dccvalidator.Rmd`~

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dccvalidator
 Title: Metadata Validation for Data Coordinating Centers
-Version: 0.3.0.9004
+Version: 0.3.0.9005
 Authors@R:
     c(person(given = "Kara",
              family = "Woo",
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@
 - Fix bug in Data Summary information boxes to not count NA
 - `check_all()` now checks the biospecimen metadata for ages over 90 in the
   `samplingAge` column
+- Results boxes now contain explanations of their contents
 
 # dccvalidator v0.3.0
 
--- a/R/app-ui.R
+++ b/R/app-ui.R
@@ -144,7 +144,7 @@ app_ui <- function(request) {
                     results_boxes_ui("Validation Results")
                   ),
                   tabPanel(
-                    "Data summary",
+                    "Data Summary",
                     fluidRow(
                       shinydashboard::box(
                         title = "Dataset summary",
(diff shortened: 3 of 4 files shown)


karawoo:
I like the idea of pointing them to the data summary tab. In that case, I think the info should only appear after the results have been generated, since otherwise the failures box will be un-collapsed and will tell people to look at a summary that doesn't exist yet.

Aryllen:
> ...will tell people to look at a summary that doesn't exist yet.

I don't think that will be too confusing due to how this is worded already. It sounds like "if there are failures here, do this," versus "there are failures, do this."

Potential suggestion:

> Any failures in this box must be corrected. A summary of data uploaded can be found in the Data Summary tab, and may help discover the reason behind the failures. Contact the DCC team if you have any questions.

karawoo:
Good point, I guess it's not too confusing (or if it is, we'll hear about it and can change it).

We talked about moving the data summary section to the same tab as the validation results (#377). Do you think we will do that? We'll just have to remember to change the wording if so. Or, maybe if we tell people to look to the data summary tab to better understand their failed results, that will also be enough to address #377 too?

Aryllen:
> We talked about moving the data summary section to the same tab as the validation results (#377). Do you think we will do that? We'll just have to remember to change the wording if so. Or, maybe if we tell people to look to the data summary tab to better understand their failed results, that will also be enough to address #377 too?

I'm concerned about whether the page will be too long/cumbersome with the summary section below. I have not tried it out, yet. I do like your idea that it might be enough to simply tell people that it's there. I still think we might need to add at least a little commentary on that tab, however. So the failures box pointing to the tab might be part 1 of #377 (get them to look at it), but part 2 would be giving more information on what the purpose of the sections are.

karawoo:
Sounds good to me. I added your suggested language and fixed the capitalization (technically a separate change from this PR, but small enough I hope it's ok).

Aryllen (approved):
Wonderful! Thank you!

PR #409, opened 2020-09-04
Hover tips

### LINKED ISSUE #372: Add tips for each file upload
Add a question mark tip for each file upload that someone can click on and get more information/instruction about.

Based on feedback from Duminda (putting instructions closer to the task).

### LINKED ISSUE #373: Look into closing of information boxes
The information boxes accessed via the question mark only go away if the question mark is clicked again. @karawoo looked into this before when designing the app and did not find anything. Check into this, again, to see if this feature has been added in the meantime (or could be easily implemented some other way).

Aryllen:
Fixes #372, fixes #373

Draft because of the note and questions below. Also probably could have made this two PRs instead of one...

- Changes behavior of the information tips to appear with hover instead of click
  - Note: need to figure out why some tips will not appear. Noticed this for `check_ages_over_90` and `check_parent_syn`.
- Adds information tips when hovering over file inputs for both documentation and validation sections
  - Questions: Should this tip appear when hovering over the `fileInput` or should I add a question mark bubble for the tip (i.e. is this too obnoxious as it is)? How is the wording for the tips?

~~- [ ] Added tests for new functions or for new behavior in existing functions~~
~~- [ ] If adding a new exported function, added it to the reference section of `_pkgdown.yml`~~
~~- [ ] If adding a new configuration option, documented it in `vignettes/customizing_dccvalidator.Rmd`~~

### DIFF
--- a/R/app-ui.R
+++ b/R/app-ui.R
@@ -126,7 +126,7 @@ app_ui <- function(request) {
                     )
                   ),
                   "Information",
-                  "Select the assay metadata file. This file should have one row per specimen, with data about the assay performed on each specimen in the experiment. If adding a new dataset to an existing dataset, please include all previous assay specimens.", # nolint
+                  "Select the assay metadata file. Depending on the assay, this file should have one row per specimen or one row per individual (indicated in the template), with data about the assay performed on each specimen or individual in the experiment. If adding a new dataset to an existing dataset, please include all previous assay specimens or individuals. Please be sure to choose the correct assay type from the drop-down above, as well.", # nolint
                   placement = "top",
                   trigger = "hover"
                 ),
--- a/R/app-ui.R
+++ b/R/app-ui.R
@@ -71,10 +71,32 @@ app_ui <- function(request) {
                 # Files to be validated
                 conditionalPanel(
                   condition = "input.species != 'drosophila'",
+
+                  popify(
+                    shinyjs::disabled(
+                      fileInput(
+                        "indiv_meta",
+                        "Individual metadata file (.csv)",
+                        width = NULL,
+                        accept = c(
+                          "text/csv",
+                          "text/comma-separated-values,text/plain",
+                          ".csv"
+                        )
+                      )
+                    ),
+                    "Information",
+                    "Select the individual metadata file. This file should have one row per individual, with data about each individual in the experiment. If adding a new dataset to an existing dataset, please include all previous individuals.", # nolint
+                    placement = "top",
+                    trigger = "hover"
+                  )
+                ),
+
+                popify(
                   shinyjs::disabled(
                     fileInput(
-                      "indiv_meta",
-                      "Individual metadata file (.csv)",
+                      "biosp_meta",
+                      "Biospecimen metadata file (.csv)",
                       width = NULL,
                       accept = c(
                         "text/csv",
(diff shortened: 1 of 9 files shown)


R/app-ui.R
```diff
@@ -71,57 +71,83 @@ app_ui <- function(request) {
...
+                      accept = c(
+                        "text/csv",
+                        "text/comma-separated-values,text/plain",
+                        ".csv"
+                      )
                     )
-                  )
+                  ),
+                  "Information",
+                  "Select the assay metadata file. This file should have one row per specimen, with data about the assay performed on each specimen in the experiment. If adding a new dataset to an existing dataset, please include all previous assay specimens.", # nolint
```
karawoo, line 129: "one row per specimen" doesn't seem like it will always be the case (e.g. for behavioral assays)

R/app-ui.R
```diff
@@ -71,57 +71,83 @@ app_ui <- function(request) {
...
+                      accept = c(
+                        "text/tsv",
+                        "text/tab-separated-values,text/plain",
+                        ".tsv"
+                      )
                     )
-                  )
+                  ),
+                  "Information",
+                  "Select the manifest file. This file should have one row file to be uploaded to Synapse, including the metadata files, with data about the contents of each file, as well as the study itself. The manifest will be used to upload the data.", # nolint
```
karawoo, line 148:
```suggestion
                  "Select the manifest file. This file should have one row per file to be uploaded to Synapse, including the metadata files, with data about the contents of each file, as well as the study itself. The manifest will be used to upload the data.", # nolint
```

karawoo (commented):
I think hovering over the question marks in the results is really nice! I do think it would be good to have question marks for the file inputs and have the tooltips tied to those. I think the wording is pretty good, other than a few small comments. One thing is this might also be a nice place to link to example files (#311). But I don't know how well that will work, because right now if you mouse over the tooltip bubble it disappears, so you could never click on a link.

R/app-ui.R
Aryllen: That's a good point. Hmm. Do you have a suggestion on how to clarify this better?

Aryllen:
I can move the tips to the question mark bubbles. I like the idea of adding links to example files, but it definitely would be difficult with the hover view. I could make specifically those ones not hover-able. We could potentially put a link next to or below the fileInput that takes them to an example, as well.

R/app-ui.R
karawoo: Maybe "depending on the assay, this will have one row per specimen or one row per individual (the template will indicate which"

Aryllen:
Moving the tooltip to a question mark circle is simple, but placement is rough.

Tooltip for indiv is placed above the fileInput (image below). Tooltip for bio is placed below it (ends up above assay).


Tooltip for indiv left as it was above, but tooltip for bio placed first in a fluidRow with the fileInput second (image below). Looks nice for the most part, even though the fileInput has to shrink to make room.

However...when you hover (or click if we had click-action), the tooltip info box is long and skinny. Not sure how to fix that.


I'm pretty much against putting the question mark circle below since it's unruly. However, which of the options do you think would look best, @karawoo?
- question mark circle above
- question mark circle to the left of fileInput
- no question mark circle and have tooltip pop up when cursor is anywhere over fileInput (i.e. how it is on this PR currently)

karawoo:
I definitely don't like putting it below, it looks like it goes with the wrong input.

I think of these options I like left best. Did you consider putting it on the right? That seems the most natural to me but maybe it doesn't work. Overall I think any of them except below would be ok.

Aryllen:
> I think of these options I like left best. Did you consider putting it on the right?

I actually tried that before trying the left one. It still has the weird skinny popup text, but it has the problem that it becomes unruly if the screen size is off. It tries to put it below. 😒 It actually looks the nicest, though (minus the skinny text issue).


If the window is too small, it pops below so it's above assay.


If we aren't worried about people using the app in teeny, tiny windows, we could put it to the right side.

Also, I found the problem with why some of the behaviors were not appearing: they had single quotes in them, which breaks the html returned from `report_results()`. Changed to backticks since those look similar anyway.

karawoo:
Aaaah I see. That is not great. Probably there is something we could do with CSS to improve things, but maybe that's too much of a rabbit hole.

Regarding the behaviors not showing up -- should we escape the single quotes in the behavior text?

Aryllen:
@karawoo, there is a weird flickering with the information boxes on fileInputs when I have it attached to the question mark circle. It's super annoying. I didn't notice until I was doing more testing than a quick glance. I had assumed the app was still loading, but it persists long after the app is open and doesn't depend on whether the pop-up has been opened before. The flickering is worse with the question-marks on the left. I think it is trying to figure out how to size the box since the fluidRow confines it to a column width, and a rather small width at that (it looks silly with width > 1 unit).

I am opening this for review, but it still has the hover tooltip tied directly to the fileInput. If this is not acceptable, I can keep looking into tying it to the question-mark circle. I think putting the tooltip above the fileInput would be a sure-fire way to remove the flickering, but it still looks a little weird.

Aryllen:
The `hover-tips-top` branch shows what it would look like to move them to the question mark circles above the fileInputs.

R/check-parent-syn.R
```diff
@@ -27,7 +27,7 @@ check_parent_syn <- function(manifest,
       )
     )
   }
-  behavior <- "Synapse IDs should follow the format 'syn' followed by digits"
+  behavior <- as.character("Synapse IDs should follow the format &#39;syn&#39; followed by digits") # nolint
```
karawoo, line 30: Is `as.character()` needed here? It's not present for the other behavior message you edited.

karawoo (commented):
I think I've managed to have the ❓ icon to the right and have it stay properly in place if the window is resized by putting the whole input + icon in a `<div>` and reusing some of the CSS we use for the results. I also switched the tooltip placement to the left, which seems to avoid the skinny box issue.

```r
div(
  class = "result",
  div(
    class = "wide",
    shinyjs::disabled(
      fileInput(
        "indiv_meta",
        "Individual metadata file (.csv)",
        width = NULL,
        accept = c(
          "text/csv",
          "text/comma-separated-values,text/plain",
          ".csv"
        )
      )
    )
  ),
  popify(
    tags$a(icon(name = "question-circle"), href = "#"),
    "Information",
    "Select the individual metadata file. This file should have one row per individual, with data about each individual in the experiment. If adding a new dataset to an existing dataset, please include all previous individuals.", # nolint
    placement = "left",
    trigger = "hover"
  )
)
```

Will you try this and let me know what you think? If it doesn't work, I'll approve this PR as-is, but I do think it would be nice to give the tooltips a little more chill by confining them to an icon.

R/check-parent-syn.R
Aryllen: Nope, I was throwing everything at that behavior to try to get it to behave nicely. After discovering the problem with the single quotes, I forgot to remove that bit. 🤦‍♀️

Aryllen:
That works great, @karawoo! I didn't even think to try dividing the space up with `div`s. 🤦‍♀️ I think one of the big problems was trying to get it to work with `fluidRow`, which has less flexibility (or rather almost too much flexibility and not enough control). I merged `hover-tips-right` into this branch and am updating the documentation section to have the same behavior.

Aryllen:
So that was fun...I kind of went crazy with the hover tips and put them on all the things. Let me know if it's too much. 😂

karawoo (approved):
This seems great, just needs to resolve the merge conflict but otherwise I'm 👍

Aryllen:
Thanks, @karawoo!

PR #417, opened 2020-10-06
Note differenece in upload speeds between clients in instructions

kelshmo:
This PR adds a single clarifying sentence to the instructions - the Python client can provide faster upload speeds.


Aryllen (approved):
Great!

Aryllen:
I just noticed that it is actually failing on the GH Actions. I'll check that out. It shouldn't be...

PR #420, opened 2020-10-08
Remove shiny file input progress bar

### LINKED ISSUE #151: Change 'Upload Complete' indicator on fileInputs
The `fileInput` gives a status that shows `Upload Complete` after giving it a file. This only uploads to a temp file on the server, though. The `Submit` or `Validate` button is what uploads the data to Synapse. By having the `Upload Complete` on the `fileInput`, it makes it look like the user does not need to press the `Submit` button in Documentation. It might also be a little confusing in the Validator.

https://stackoverflow.com/questions/40620176/getting-rid-of-the-status-bar-in-file-upload-in-shiny

### LINKED ISSUE #413: Remove upload progress bar on fileInputs if possible
Some people are thinking that they are done when they see "upload complete" on the progress bar for the fileInputs. It would be better if we didn't have that, but not sure how simple that would be.

karawoo:
Fixes #413, fixes #151

- Removes fileInput progress bar completely. I'm on the fence about how this looks, as I kind of like the visual feedback of the progress bar, but if it's confusing to people I'm fine with removing it. After this PR is merged we will need to update the version of dccvalidator tracked in the lockfile, since I think it's the installed version whose CSS file gets used when the app is deployed. Relatedly, if you want to test this locally I think you need to run `devtools::install()` before running the app.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dccvalidator
 Title: Metadata Validation for Data Coordinating Centers
-Version: 0.3.0.9009
+Version: 0.3.0.9010
 Authors@R:
     c(person(given = "Kara",
              family = "Woo",
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,7 @@
 - All tooltips now pop-up with hover instead of clicking
 - Add hover tooltips to all fileInputs
 - Fix bug in behavior for `check_ages_over_90()` and `check_parent_syn()`
+- Remove progress bars for file input boxes
 
 # dccvalidator v0.3.0
 
--- a/inst/app/www/custom.css
+++ b/inst/app/www/custom.css
@@ -49,3 +49,7 @@ img {
   background-color: #367FA8;
   padding: 3%;
 }
+
+.shiny-file-input-progress {
+    display: none
+}


Aryllen:
I also like the feedback of the progress bar, but it is definitely a problem. One good thing is it does still show that a file was selected by displaying the path.

Should we get a third opinion on removing the progress bar?

Aryllen (commented):
Let's go ahead with removing the progress bars. My only thought is whether we should also add a check that the number of files expected is the number of files uploaded to the server. By that, I mean that we check that the files selected have finished uploading. I am not sure of the best way to do that. I _think_ the fileInput's `name` should be available before the file finishes uploading. So for every fileInput that has a `name`, there should also be a valid fileInput `datapath`.

Technically, the files upload relatively quickly so people should not be clicking the validate or submit button before the files are on the server, but the possibility does exist that one or more of the files would still be in the process of uploading.

karawoo:
> Technically, the files upload relatively quickly so people should not be clicking the validate or submit button before the files are on the server, but the possibility does exist that one or more of the files would still be in the process of uploading.

Yeah, this is a concern of mine as well. The upload is fast on a good internet connection, but that may not always be the case. It's possible someone might click the validate button and not all the data would be included in their results. But I suppose we can see if that becomes a problem.

Aryllen:
> But I suppose we can see if that becomes a problem.

This might be what we will have to do for now. I've been trying to think of a way to test/check that the files are finished uploading, but I cannot think of how at the moment. Trying locally is hard since it's not actually uploading and my internet is too fast to try to do a test with the server version. I'll approve this and make an issue for tracking any problems and ideas for solutions that come up.

Aryllen (approved):

PR #421, opened 2020-10-09
Make attempt_login() return NULL if login errors

### LINKED ISSUE #419: Troubleshoot check for being logged in when on CI
The GH Actions were updated recently causing a problem with the environment variables not being accessed in the CI. Since the Synapse client should not have been able to log into Synapse, the tests that hit Synapse should have been skipped. However, they were not. Instead, it still tried to pull from Synapse and we ended up with 401 errors. The login problem is currently fixed, but functions surrounding skipping of tests should be made more robust.

karawoo:
Fixes #419

- Previously `attempt_login()` would return the `try` object if login failed, which is not what we want. This change checks the result of the attempted login and returns `NULL` if it fails. 

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -1,4 +1,4 @@
-## Check if we are running on travis
+## Check if we are running on CI
 on_ci <- function() {
   if (identical(Sys.getenv("CI"), "true")) {
     return(TRUE)
@@ -22,7 +22,12 @@ attempt_instantiate <- function() {
 ## within Sage, do nothing.
 attempt_login <- function(syn, ...) {
   if (on_ci() & !is.null(syn)) {
-    try(syn$login(), silent = TRUE)
+    attempt <- try(syn$login(), silent = TRUE)
+    if (inherits(attempt, "try-error")) {
+      return(NULL)
+    } else {
+      return(syn)
+    }
   } else if (reticulate::py_module_available("synapseclient") & !is.null(syn)) {
     syn$login(...)
   } else {


Aryllen:
I was wondering if we should add some tests with mocking, but I am not sure if we can since we can't mock the client itself. However, that train of thought brought me to whether we did have tests along these lines.
https://github.com/Sage-Bionetworks/dccvalidator/blob/9ea44b06f937ce2a213fd46cde12e2210749d1cc/tests/testthat/test-utils.R#L6-L25

The second test should have failed because it's on CI (i.e. should not be skipped) and I removed `syn_ci_login()` since GH Actions isn't using environment variables to get the login info. So the second test needs fixing because it's using a function that no longer exists, but I think our `on_ci()` function isn't working how we want it to, either.

karawoo:
I think that test is skipping because it's trying to get a Travis environment variable that doesn't exist because we're no longer using Travis, and forgot to change this test when we switched to GitHub Actions.
https://github.com/Sage-Bionetworks/dccvalidator/blob/9ea44b06f937ce2a213fd46cde12e2210749d1cc/tests/testthat/test-utils.R#L17-L21

karawoo:
Looking at this again, I think this change isn't actually what's needed. The return value doesn't matter so much because it doesn't typically get assigned to anything; the function is being called for its side effects on the `syn` object. I'm going to need to rethink this, sorry for the confusion!

Aryllen:
> Looking at this again, I think this change isn't actually what's needed. The return value doesn't matter so much because it doesn't typically get assigned to anything; the function is being called for its side effects on the `syn` object. I'm going to need to rethink this, sorry for the confusion!

That's a good point! Thank you for working on this.

karawoo:
I'm going to close this and think more about it. Possibly in the case with kelsey's PR, `logged_in()` was returning `TRUE`, maybe with an empty string as the username (because it couldn't find the environment variable)? So maybe that's where the change needs to happen.

PR #424, opened 2020-10-13
Add loading screen with waiter

### LINKED ISSUE #422: Add loading screen
Add the fancy loading screen instead of making people see buttons they can't click on until the login dance completes: https://github.com/Sage-Bionetworks/SynapseShinyApp/pull/15

karawoo:
Fixes #422 

- Adds a loading screen using waiter. Note that with this PR again you may need to run `devtools::install()` locally before it will work.


Aryllen (changes requested):
I just checked this out. I did not have to use `devtools::install()` to get it to work. However, I was also using Rstudio and in the dccvalidator 'project'. It might be different for other IDEs.

Only issue I saw was that it loaded a portion of the app UI for a brief moment, then switched to the loading screen. This may need to be called elsewhere. I also have not yet seen what happens if I am not logged in. However, it looks like it should just give a message. This is great! If they aren't logged in, they don't see the app. In that case, we can remove all the extra `shinyjs::enable/disable`. However, I also think it would be helpful to give them a link to the vignette for 'Using the App' since they do not need to be logged in (currently) to just see the instructions of how to use the app, what's expected, etc.

Aryllen (changes requested):
Checked this out on my test version on the Shiny server. I don't have the same problems as I do locally (where it tries to show the app, then switches to the loading screen).

I would still like to discuss:
- if I am not logged in, the loading screen just continues saying that it's trying to connect to Synapse. It isn't seeing that I am not signed in and giving me any extra message.
- removing the unnecessary `shinyjs::enable/disable` since they should not see the app without being logged in.
- adding a link in the 'not signed in' message to documentation they can look at if they are not logged in.

karawoo:
@Aryllen will you send me the link to your test version that's live?

karawoo:
Actually, I can reproduce this with the amp-ad-metadata-dictionary app. I think the problem is that when you run it locally you get an error if login fails, but if the session token is missing it doesn't throw an error. I need to report this back to the synapseShinyApp repo and come up with a fix.

karawoo:
I cherry picked out the github actions updates into #445; this PR will fail on github actions until that one is merged

Aryllen:
Note to self: need to fix conflict and check that this works every time (the previous version failed sometimes).

PR #433, opened 2020-11-07
Add ability to update metadata template dictionaries

### LINKED ISSUE #430: Add functionality for updating template dictionary
The templates are Excel files with two extra tabs, one for the key definitions and one for the accepted key-value pairs. There is a [script in the AD-DCC repo](https://github.com/Sage-Bionetworks/AD-DCC/blob/master/add_template_description_sheets.R) to do this. We should incorporate the functionality from this into dccvalidator, along with documentation. Additionally, we could use GH Actions to update the templates on a regular basis.

Steps to complete:
- [ ] Add necessary functions
- [ ] Add appropriate function testing
- [ ] Add any needed config parameters
- [ ] Add documentation for new feature
- [ ] Create GH Action to auto-update templates

Aryllen:
Fixes #430

- Add script in /inst/scripts folder that will update templates based on the config chosen. Added README for how to use.
- Add an exported main function, `update_template_dictionaries`, that takes in the annotations and synIDs for templates, downloads the templates, and adds the new dictionary sheets to each. Sends back the Synapse File object.
- Add an exported function for verifying that the annotations structure is valid (i.e. no key has >1 definition or columnType).
- Add an exported function for adding the dictionary template sheets.
- Add an exported function for pulling the template synIDs from the config and putting them in a vector.

- Add helper functions for:
  - generating the key definition sheet. This was a syndccutils function. I just rewrote the function to get what we needed. I did not want to add a non-CRAN dependency to dccvalidator.

- update `attempt_login` to pass back the synapse client object. This was making some of my tests fail because `syn$synapse()` will pass back `NULL` after logging in.
- Add tests for all of the new functions.
- Add documentation for new functions.
- Attempted to update dependencies with renv. Fingers crossed it works on Mac. It does seem to be working, at the moment, within my docker container.


Aryllen:
Well those errors are fun. I did fix the problems with forgetting to import the pipe and not having visible binding for variables used in dplyr functions. My tests run fine locally if I run them one by one. However, some tests are failing if I run with `devtools::check()` (when I initially checked they were, but I probably should have cleared my system cache first). I will tackle this next week.

Aryllen:
I think I found the main problems with the tests.

1. There are data frames being used in tests and, for the non-latest release versions of R, are changing strings to factors. This would be a reason why, previously, the tests succeeded on some (release versions) and failed on others.
2. Locally, when I run the tests, it runs the tests against the package level config.yml, not the tests/config.yml. On GH, I believe it runs the tests against tests/config.yml. Since default is not the same set of synIDs for the templates in both configs, the tests fail locally if I expect tests/ level synIDs in `get_template_synIDs` and on GH if I expect package level synIDs. A fix would be to use the same set of synIDs in both configs for 'default'. I am still wrapping my head around why this would be the case, in general, though.

Current extra problems not related to the tests: update_template_dictionaries() is not functioning because I'm a doofus. Working on fixing this and will update the other problems at the same time...

Aryllen:
Yay, the tests all passed. Still need to test my script locally before pushing. I may put the GH Action to run it regularly as a separate issue.

Aryllen:
@karawoo, assuming these checks all pass, I think this PR is ready to review, again. Let me know if you have any questions.

inst/scripts/README.md
```diff
@@ -0,0 +1,52 @@
...
+This script updates the dictionary sheets in existing metadata template files. A metadata template is expected to be an Excel (.xlsx) file with 3 sheets:
+- template: header of metadata keys
+- dictionary: definitions of the metadata keys present in the template (columns: key, description)
+- values: enumerated values for the metadata keys present in the template (columns: key, value, valueDescription, source)
+
+### Requirements
+
+#### System
+
+This script requires the dccvalidator package (version >= 0.3.0.9012). Install the latest version with:
```
karawoo, line 14: Should this be 0.3.0.9011 (the current version after this PR)?

DESCRIPTION
```diff
@@ -31,12 +31,15 @@ Depends:
...
     config,
+    dplyr,
     ggplot2,
     glue,
     golem,
     htmltools,
     knitr,
+    magrittr,
     markdown,
+    pkgload,
```
karawoo, line 42: I don't see where this is used, and R CMD check produces a `NOTE` saying that it's not used. I think we can remove

renv.lock
```diff
@@ -1,971 +1,958 @@
...
-      "RemoteType": "github",
-      "RemoteHost": "api.github.com",
-      "RemoteUsername": "Sage-Bionetworks",
-      "RemoteRepo": "dccvalidator",
-      "RemoteRef": "master",
-      "RemoteSha": "531375ccbc2e323d2c2f5bc4d70b02880c6b362d",
-      "Hash": "0efbac7c78190be952b066c0a4bcc47b"
+      "Version": "0.3.0.9010",
+      "Source": "Repository",
+      "Repository": "CRAN",
```
karawoo, line 205: Something is wrong here; dccvalidator 0.3.0.9010 is not on CRAN, so this fails to install when I do `renv::restore()`

karawoo (changes requested):
I'm not able to get the tests to pass locally, but I see they pass on CI. Before digging into it too much further I'd like to sort out the issue with the lockfile I've mentioned below so I can get the same packages installed and see if that affects things.
 
One other comment though is that this brings us above the CRAN-recommended number of packages in Imports, and R CMD check produces a NOTE about that. The next time submit to CRAN we will need to justify why we need so many packages; it might be a good idea to move some of them to Suggests now and avoid having to deal with it later.

inst/scripts/README.md
Aryllen: The DESCRIPTION actually needs to be updated in this PR. #435 was merged first so this one needs to be bumped to 3.0.9012. I just pushed the change for this.

DESCRIPTION
Aryllen: The note was a little annoying so I added it in. 😆 But it's a good point that it's not needed for the package itself so users shouldn't have to have it. I went ahead and removed for now.

renv.lock
Aryllen: Sorry! Fixed.

R/metadata-template-dictionary.R
```diff
@@ -0,0 +1,315 @@
...
+#'     "First name of individual",
+#'     "Last name of individual"
+#'   ),
+#'   columnType = c("string", "string")
+#' )
+#' verify_dictionary_structure(dat)
+#' }
+verify_dictionary_structure <- function(dictionary) {
+  # Check that dictionary exists and has needed elements
+  if (is.null(dictionary) || is.na(dictionary)) {
```
karawoo, line 103: We're seeing an error in the example here that includes the message

```
----------- FAILURE REPORT -------------- 
   --- failure: length > 1 in coercion to logical ---
   --- srcref --- 
```

I think the `is.na(dictionary)` here is what's causing the issue. Below is what `is.na()` returns for the data frame in the example:

``` r
dat1 <- data.frame(
  key = c("first_name", "first_name"),
  description = c(
    "First name of individual",
    "Last name of individual"
  ),
  columnType = c("string", "string")
)
is.na(dat1)
#>        key description columnType
#> [1,] FALSE       FALSE      FALSE
#> [2,] FALSE       FALSE      FALSE
```

However the data frame could be all NAs and would fail this check even if it had the right columns:

```r
dat2 <- data.frame(
  key = c(NA, NA),
  description = c(NA, NA),
  columnType = c(NA, NA)
)
is.na(dat2)
#>       key description columnType
#> [1,] TRUE        TRUE       TRUE
#> [2,] TRUE        TRUE       TRUE

is.null(dat2) || is.na(dat2)
#> [1] TRUE
```

This check seems redundant since if someone explicitly calls `verify_dictionary_structure(NULL)` or `verify_dictionary_structure(NA)`, the next line would catch it since they're `NULL` and `NA` aren't data frames. So I think we could take out the first if block all together.

Aryllen: Bah, you are totally right. I left this in after adding the next line since it was a better check, but didn't remove and didn't associate the error message with that part. 🤦 Leaving in extraneous things (and not "seeing" them) seems to be my big downfall. 😆 Going to have to use the old 'essay' method of reading things out loud. Haha.

Aryllen:
> One other comment though is that this brings us above the CRAN-recommended number of packages in Imports, and R CMD check produces a NOTE about that. The next time submit to CRAN we will need to justify why we need so many packages; it might be a good idea to move some of them to Suggests now and avoid having to deal with it later.

That's a good point. I moved the added packages to suggests. They are needed for the tests so I have them being loaded in testthat.R. Would it be better practice to just load those libraries in the specific test file for them?

Aryllen:
Currently passing locally. I did have to move magrittr back to Imports. Only warning is the one below:

> > checking for unstated dependencies in examples ... OK
>    WARNING
>   ‘qpdf’ is needed for checks on size reduction of PDFs

inst/scripts/update-metadata-template-dictionaries.R
```diff
@@ -0,0 +1,104 @@
...
+  )
+}
+
+## Get annotations
+# Get all annotations
+annots <- purrr::map_dfr(
+  config::get("annotations_table"),
+  get_synapse_annotations,
+  syn = syn
+)
```
Aryllen, line 72: @karawoo, I thought of this based on our discussion about the two sets of annotation tables potentially having a duplicate. I could add `dplyr::distinct()` here to remove duplicate rows before continuing on; this would stop it from printing the key/values twice. The verification in the next step isn't a problem; it won't fail if there are duplicate rows, only if the columnTypes or descriptions for the same key are not identical.

However, my question is if this _should_ be a separate function that is potentially called in both the shiny app and this script. The function would just return the annotations without duplicated rows. Not sure if a function for this is overkill, though, especially since a cursory look/test with dccvalidator shows that the validator functions should be fine with duplicate rows.

karawoo (approved):
This looks good to me and all tests pass locally now 🎉 

Adding `distinct()` to the place you mentioned seems fine and, as we discussed in our 1:1, probably doesn't need to be a separate function since it wouldn't need to be reused.

We can merge this as-is, or if you want to make that one change first that's fine too.

PR #441, opened 2020-12-21
Update system requirements installation in GHA

karawoo:
Hopeful that this will fix the builds for #424 -- usethis now uses gert, which has some additional system requirements. I'm hopeful this approach will be more robust in the future than adding the dependencies manually.


karawoo:
Ugh

PR #444, opened 2020-12-29
Update column types

### LINKED ISSUE #443: New columnTypes in synapse annotations table break dccvalidator
The Synapse annotations table is now using JSON Schema's types (`"string"`, `"number"` etc.). This breaks dccvalidator, which expects the old columnTypes: 
https://github.com/Sage-Bionetworks/dccvalidator/blob/70ff704d8b663a8235f009c193009b239230cf8e/R/check-annotation-values.R#L256-L262

This is a breaking change and a fix is forthcoming. Both the AMP-AD and PEC dccvalidator instances will need to be updated. We could also change the annotation table values instead/in addition to updating dccvalidator.

karawoo:
Fixes #443 

- Fixes `check_type()` to support JSON Schema types

- [X] Added tests for new functions or for new behavior in existing functions

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dccvalidator
 Title: Metadata Validation for Data Coordinating Centers
-Version: 0.3.0.9012
+Version: 0.3.0.9013
 Authors@R:
     c(person(given = "Kara",
              family = "Woo",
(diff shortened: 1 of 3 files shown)


karawoo:
I am going to merge this since it is an issue that is currently affecting the production version of the app. I didn't think anyone was likely using it right now, but we've gotten an email from someone who wants to join the AD contributor team, so maybe they are. We can always roll this back later or change it if needed.

Aryllen:
This works! Thank you for getting this taken care of.

I think we should probably change the types in the annotations table, as well, which won't take long. I will put that on my to do list.

PR #446, opened 2020-12-29
Modularize validation

### LINKED ISSUE #220: Modularize the validation section of the app
Similar to how we do the documentation page

karawoo:
Fixes #220 

- This is something I've wanted to do for a while: break out the validation logic into a module, so that the main server and UI functions are much smaller. I've tested this out and (with the additional changes in #444) it seems to work locally, but we should test it out thoroughly because a lot of code has moved around.

I have not updated to the newer way of calling modules in shiny -- I'm still using `callModule()` since that's what we use elsewhere. Later we may want to update everything to the new way.


Aryllen (approved):
Tested in a brand new directory in the server. Tested locally, again, as well (removing the accessToken part of the login or else it logs me in anonymously). It appears to be working fine and the tests all passed. Merging! 🤞🏻

PR #448, opened 2020-12-29
Empty rows

### LINKED ISSUE #428: Remove empty rows from data
There has been a problem recently with metadata files having extra, empty rows, which then gives failures in dccvalidator (incomplete required columns, for example). This isn't necessarily a bug, but it does make for undesired behavior. I don't know if we can really require people to know how to remove the extra rows that were most likely added by accident when converting the Excel template to csv. Instead, we should just remove the empty rows ourselves before doing the checks.

karawoo:
Fixes #428 
Depends on #446 

- Adds a new function to filter out empty rows and applies it to all uploaded files before validating.

- [X] Added tests for new functions or for new behavior in existing functions

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dccvalidator
 Title: Metadata Validation for Data Coordinating Centers
-Version: 0.3.0.9013
+Version: 0.3.0.9014
 Authors@R:
     c(person(given = "Kara",
              family = "Woo",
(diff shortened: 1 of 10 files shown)


karawoo:
(This branch will need to be rebased after #446 is merged)

Aryllen (changes requested):
This doesn't work if columns are different types. Will update when I get a chance.

PR #449, opened 2020-12-30
Lint, style, and regenerate docs

### LINKED ISSUE #439: Lint, style, and docs check
There are some lint, style, and docs that get updated when I run these for a PR. I don't include them since they are irrelevant to the change being done. However, should go through and do a check over all and fix the issues.

On a side note, our GH action for linting doesn't fail if there are linting errors. While it's annoying to have a test fail for these issues, it may be worth considering an update to have it fail. As it is, the warnings are significantly less noticeable than they were when we were using Travis.

karawoo:
Fixes #439 

- A few changes to code style to satisfy lintr. I re-ran `styler::style_pkg()` which also introduced some formatting changes. And, I've regenerated the .Rd files that were out of date.

### DIFF
--- a/R/app-server.R
+++ b/R/app-server.R
@@ -33,8 +33,7 @@ app_server <- function(input, output, session) {
 
   observeEvent(input$cookie, {
     is_logged_in <- FALSE
-    tryCatch(
-      {
+    tryCatch({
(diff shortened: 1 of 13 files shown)


Aryllen (approved):
Thank you for doing this!

PR #451, opened 2021-01-06
Update test for logging in on GH Actions

### LINKED ISSUE #419: Troubleshoot check for being logged in when on CI

Aryllen:
Fixes #419. Based on my memory and checking through the references on this, I believe this issue will be resolved.

- Fix comments and test names referring to Travis, specifically, instead of CI.
- Update github action for R-CMD-Check to write the .synapseConfig file to the home directory, instead of tmp, for all checks except Windows, and add missing bracket to the Windows .synapseConfig.
- Update the test for logging in to work on GH Actions.

Remaining before merge:
- [x] Verify that this fixes the test for logging in while on GH
- [x] Increment version
 
### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -98,15 +98,15 @@ jobs:
           OUTPUT_FILE=~/.synapseConfig
           cat > "$OUTPUT_FILE" << EOM
           [authentication]
-          username = "${{ secrets.SYNAPSE_USERNAME }}"
-          apikey = "${{ secrets.SYNAPSE_API_KEY}}"
+          username = "${{ secrets.SYNAPSE_USER }}"
+          apikey = "${{ secrets.SYNAPSE_APIKEY}}"
           EOM
           chmod +x "$OUTPUT_FILE"
           
       - name: Setup Synapse config file for testing (Windows)
         if: runner.os == 'Windows'
         run: |
-          "[authentication]`nusername = ${{ secrets.SYNAPSE_USERNAME }}`napikey = ${{ secrets.SYNAPSE_API_KEY}}" | Out-File -FilePath ".synapseConfig"
+          "[authentication]`nusername = ${{ secrets.SYNAPSE_USER }}`napikey = ${{ secrets.SYNAPSE_APIKEY}}" | Out-File -FilePath ".synapseConfig"
 
       - name: Check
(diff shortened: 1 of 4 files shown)


Aryllen:
So this is interesting. Apparently it was fine to write the .synapseConfig to tmp instead of home since tests are now failing for not being logged in. Technically these shouldn't fail if not logged in, but rather skipped.

karawoo:
I think the .synapseConfig file is looking for credentials in the wrong environment variable names. The repo has secrets called `SYNAPSE_APIKEY` and `SYNAPSE_USER`.

Screen Shot 2021-01-06 at 10 58 03 AM

The credentials file we create in the gha workflow looks for `SYNAPSE_API_KEY` and `SYNAPSE_USERNAME`.

Aryllen:
Thank you for catching that, @karawoo! It looks like it's working now, but I am still concerned whether this is working correctly.

Before this PR, the tests that _should_ be skipped if not logged in were passing (not this test in this repo, but rather the other tests that touch Synapse). I assume they were being skipped for not being logged in.

After I moved the .synapseConfig file location for all OSs except Windows, but still had the wrong secrets, these tests were failing (they still 'passed' on Windows). Since the secrets were wrong, these should have 'passed' in the sense that they should have been skipped due to not being logged in. I'm concerned that they were not skipped then, but seem to be passing now.

Trying to think of how to test this to make sure that the tests are passing because they are running and correct, not because the tests are skipping.

karawoo:
Maybe with the failed login `syn$username` was something other than `NULL`, causing `is_logged_in()` to return `TRUE` even though we weren't actually logged in.

karawoo (approved):
This looks good to me. Agreed we need a way to better catch this sort of issue in the future. I guess this is what we get for deciding to "skip gracefully" to make PRs from forks easier...maybe there is some way to log the skipped tests so we could at least look at the logs to make sure that we're not skipping things we don't expect to be.

Aryllen:
> maybe there is some way to log the skipped tests

This was my thought, as well. I know it will do this with `devtools::test()` so there _should_ be some way to write and get a log.

PR #466, opened 2021-04-08
Update check value

### LINKED ISSUE #207: Support validating multiple annotation values
Once multiple annotation values are officially supported in Synapse, we need to update the Synapse File and CsvFileTable methods for `check_annotation_values()` to ensure that they check each value.

Aryllen:
Fixes #207

Note: This update will only work for strings and only if the key has an enumerated set of values (free text values are ignored). If the intent is to have multiple values with a numeric or boolean, it will not check this (e.g. readLength could not have "100, 120, 150"; this would need to be 3 separate rows). Also, that's kind a bit weird so I'm currently fine with ignoring multi-annotation numeric and boolean values unless someone has good use-cases for it.

- Update `check_value()`. If a key has enumerated values in the annotation set and is a string, then the function will:
  - remove square brackets and extra quotes (needed for json-style string lists)
  - split values that are comma separated
  - trim whitespace around the values
  - check all values found against the annotation set
  - returns only the values that do not match the set
- Added tests.
- Updated a few packages in the renv lock file. If this causes problems with the checks, will revert.
- Updated NEWS.md.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,6 +13,7 @@
 - Fix bug in behavior for `check_ages_over_90()` and `check_parent_syn()`
 - Remove progress bars for file input boxes
 - Add script and functions for updating metadata template dictionary sheets.
+- Update to `check_value()`: For keys with enumerated values, parse comma-separated and json-style strings and check all values within against allowed values.
 
 # dccvalidator v0.3.0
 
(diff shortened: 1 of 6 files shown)


avanlinden:
@Aryllen This is great! The changes look good to me. I will say that for metaboliteType, some of our enumerated values contain spaces, e.g. "fatty acids". I tried a few little tests and it seems like it will be fine, though.

Juliane and I already looked through the AD metadata dictionary and we don't currently have any enumerated values with commas in them, so we will continue to enforce that in the future and reserve commas for separating string lists. Thank you!

avanlinden (approved):

Aryllen:
Thanks for checking this out, @avanlinden, and for being thorough to see if trimming space or commas would affect anything.

PR #477, opened 2021-05-14
Veoibd

cw-dvr8:
Fixes #

- This is a placeholder for adding VEOIBD to dccvalidator.  We currently don't have any metadata standards from the consortium, so I tested the Shiny app with everything pointing at PsychENCODE resources.  It seems very unlikely that we will get feedback from the consortium on the metadata before I leave, so someone will have to make modifications to this code when the standards become available.


Aryllen:
I will check this out on Monday. I see there is at least one error in the Actions, but I think it's related to setup, not the PR.

config.yml
```diff
@@ -196,3 +196,63 @@ annotations-research:
...
+  annotations_table: "syn25558026"
+  annotations_link: "https://www.synapse.org/#!Synapse:syn20729790"
+  templates_link: "https://www.synapse.org/#!Synapse:syn20729790"
+  docs_tab:
+    include_tab: TRUE
+    include_upload_widget: FALSE
+    tab_name: "Study Documentation"
+    path_to_docs_markdown: "inst/study-documentation-pec.Rmd"
+  teams:
+    - "3382930"
```
Aryllen, line 214:
```suggestion
    - "3382930"
    - "3425042"
```
This should probably have the admin team on it, as well. Also, I noticed that you, @cw-dvr8, are the only admin on that team. You may want to elevate Mette to admin.

app.R
```diff
@@ -2,7 +2,7 @@
 # To deploy, run: rsconnect::deployApp()
 # Or use the blue button on top of this file
 
-Sys.setenv(R_CONFIG_ACTIVE = "default") # Replace "default" with your config
+Sys.setenv(R_CONFIG_ACTIVE = "veoibd") # Replace "default" with your config
```
Aryllen, line 5:
```suggestion
Sys.setenv(R_CONFIG_ACTIVE = "default") # Replace "default" with your config
```
On GH, I'd like to keep this as `default`. When run it on the server, you would update this on your local version to be `veoibd`.

Aryllen (changes requested):
There's a couple changes requested.

I also noticed that the server is disconnecting on me when I visit your app. That's odd behavior. If you visit your app in the web, does it disconnect on you after a minute or so?

Aryllen (approved):
Thanks, @cw-dvr8!

Issue #292, opened 2020-02-24
Switch to OAuth


karawoo:
I have not heard the status of Synapse OAuth support lately, but if it is at a point now where users can authenticate and upload data, we should use it in place of the session token cookie. This would indirectly fix #290, hopefully.

Aryllen:
OAuth is now possible. Refer to Tom's example [here](https://github.com/Sage-Bionetworks/SynapseShinyApp/tree/update-reticulate).

PR #479, opened 2021-05-20
Log in with authToken and add staging endpoints

Aryllen:
Fixes # NA

The Synapse cookie is changing this weekend to not include a sessionToken anymore. Instead, this will be an authToken. We will be switching to oauth (#292), but this is a band-aid in the meantime.

**NOTE:** This version requires Synapse client >= 2.3.1. Upgrade your client to test. You may need to upgrade pip, first.
```
pip install --upgrade pip
pip install --upgrade synapseclient
```

- Try logging in with sessionToken. If doesn't work, log in with authToken.
- Add helper function to change to staging endpoints.
- Add 'production' config param to easily set staging endpoints for testing apps with staged version of Synapse.
- Add 'dev' config that inherits all default config params, but overwrites 'production' to be `FALSE`. Testing the app with staging endpoints would simply mean updating app.R to point to 'dev'.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dccvalidator
 Title: Metadata Validation for Data Coordinating Centers
-Version: 0.3.0.9017
+Version: 0.3.0.9018
 Authors@R:
     c(person(given = "Kara",
              family = "Woo",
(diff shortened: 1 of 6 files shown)


Aryllen:
Since this breaks this weekend and @avanlinden was able to log into an earlier test version of this fix, I'm going to merge this so that I can update dccmonitor and push changes to the server.

kelshmo:
Just want to note I also successfully tested this.

Aryllen:
Thank you, @kelshmo!

PR #481, opened 2021-05-28
Remove sessionToken backup and add check for non-anonymous username

Aryllen:
Fixes #NA

- Remove attempt to log in with sessionToken. This was in place so that the app would function through the weekend. After the Synapse site was pushed, it should have errored and tried using the authToken login. Unfortunately, it's just logging in users as 'anonymous' due to an error in the web client.
- Add a check to make sure user isn't logged in as anonymous. Currently have no reliable fix for this if they are so the modal doesn't give any instructions, just a description of the problem.

### DIFF
--- a/R/app-server.R
+++ b/R/app-server.R
@@ -37,29 +37,31 @@ app_server <- function(input, output, session) {
 
   observeEvent(input$cookie, {
     is_logged_in <- FALSE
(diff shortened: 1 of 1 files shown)


Aryllen:
Tested on Shiny server.

PR #488, opened 2021-06-08
Add OAuth client for logging into app

### LINKED ISSUE #292: Switch to OAuth

Aryllen:
Fixes #292. Note that the OAuth code used here is adapted from @thomasyu888's work on the reticulate branch of [synapseShinyApp](https://github.com/Sage-Bionetworks/SynapseShinyApp/tree/reticulate).

Test app embedded in Synapse [here](https://www.synapse.org/#!Wiki:syn25829574/ENTITY).

- Have users of app log into it via OAuth. This updates some previously existing functions to work a little differently, such as `attempt_login()`, and exports functions that were not previously exported to help with adding the feature to dccmonitor.
- Stop the app from using OAuth if running locally (i.e. interactively) and default to logging in via .synapseConfig credentials. This stops the need to add a local redirect_uri to the dccvalidator OAuth client, reducing the security risk.
- Update docs and vignettes.
- Move config.yml to inst/ and add in new(er) golem function, `get_golem_config()`. This allows for pulling in the URL from the config file easier. Also follows golem framework better. Change all references of `config::get()` to `get_golumn_config()`.
- Updated and merged in some of Kara's PRs, which would make it easier to modularize app.
- Made the main app itself a module 🤯.
- Made the OAuth a module. Note that this still requires global variables. There's a helper function for setting these up (assuming they've been globally instantiated so there's something to link to). This can be called in `.onload` to get it setup on the server side, resulting in less time for the users to wait around. I'm relatively sure these could be non-global by using a `uiOutput` in `mod_synapse_oauth_ui` and controlling it in `mod_synapse_oauth_server`. However, I need to work on other things so this is the version I'm going with for now.
- Reorganization of code.
- Fixing documentation links, and rendering of them, to open in a new tab.
- Updating the tests to work with the new config location.

Important notes:
- This change makes it much harder to spin up a new instance dccvalidator for testing on the server. 
  - The OAuth client will give back an error if the app URL is not registered. I have a test version of dccvalidator [here](https://shinypro.synapse.org/users/nkauer/dccvalidator-test/) (the URL is registered with the client).
  - The APP_URL for redirect is coming from the config file. This means that to test a different version of the Shiny app in the server, the config for that version will need to be updated to point to the test version of dccvalidator. I've tried out a few things to make this 
- The app should _appear_ to not be changed if run locally. This is because I have it default to using your local .synapseConfig file. If you do not have this file, then it should fail to log in. I could have continued using the Synapse browser cookie, but it sometimes logs in as anonymous (not desired behavior) and is ultimately a less secure way of logging in.


thomasyu888:
@Aryllen 

- Ya... implementing OAuth certainly does make it harder to test things locally.  One thing I have thought of doing is providing people (in lastpass) credentials to a oauth client that would work for shiny apps that are hosted at localhost:8100.  You can set `options(shiny.port=8100)` (or something like that to keep the port consistent). 
- You can also set multiple redirects for your OAuth client, but you would need to know the shiny server URLs before hand.  I'm sure theres probably a way to update the redirect_url's in a oauth client via rest calls.
- I like that you used `~/.synapseConfig` to log in when running this application locally.
- Is it feasible for the OAuth component to be a `golem` module?  It would be highly beneficial to just be able to install this component and use this for other applications.

R/synapse-oauth.R
```diff
@@ -0,0 +1,141 @@
...
+  synapse <<- reticulate::import("synapseclient", delay_load = TRUE)
+}
+
+## Prep OAuth client
+prep_for_oauth <- function(app_url) {
+  APP_URL <<- app_url
+  app <<- httr::oauth_app("dccvalidator",
+                         key = CLIENT_ID,
+                         secret = CLIENT_SECRET, 
+                         redirect_uri = APP_URL)
```
Aryllen, line 55: Pull in the client name, instead of hardcoding.

Aryllen (commented):
> One thing I have thought of doing is providing people (in lastpass) credentials to a oauth client that would work for shiny apps that are hosted at localhost:8100.

This could be handy! The name of the client would need to be more flexible on runtime (i.e. not hardcoded), but it would be so much simpler to have a single client that is accessible locally than having to create each individual client with this feature.

> You can also set multiple redirects for your OAuth client, but you would need to know the shiny server URLs before hand. I'm sure theres probably a way to update the redirect_url's in a oauth client via rest calls.

Yup. I have multiple redirects in our clients so I can use one for dccvalidator and one for dccmonitor. There are REST calls to update these and it's fairly simple, as long as the redirect is on the same host. I haven't messed with redirects on different hosts, yet. I'm also not sure _who_ can update these, the creator only?

> I like that you used ~/.synapseConfig to log in when running this application locally.

Thanks! It should make it easier if we do add shiny testing that needs the app itself to run.

> Is it feasible for the Synapse login (including OAuth) component to be a golem module?

I think this would be possible. I had originally discounted it because of the global vars, but I don't know that they really need to be global. Let me give it a try!

Aryllen:
Making a real PR, instead of draft since it's ready for review. The linux GH Actions are failing, I think due to the python installation.

Last things I can think of is updating the docs to reflect the latest overhaul.

Aryllen:
I've went over these changes several times myself, and @pitviper6, @kelshmo, and @avanlinden have tested a live version of the app. Going to merge this change.

PR #491, opened 2021-06-18
Add ability to use JSON schema for checking metadata keys

Aryllen:
Fixes #NA -- surprisingly don't have an issue for this that I could find. However, this is the first step towards using Synapse registered JSON schemas for metadata validation.

- Add ability to use simple (see note below) registered JSON schemas to validate that all expected metadata columns are present (`check_col_names()`)
- Update `get_template()` to accept an `id` for a registered JSON schema
- Add functions to get a validation schema from Synapse (i.e. references to other schema have been dereferenced) and pull the property names
- Some testing for the additional functions (more testing would not hurt!)
- Updating of docs, NEWS, etc
- Update default config to have the id of a registered JSON schema as an example

NOTE: This currently works for simple JSON schema. By simple, I mean that there is a top level `properties` object that has the complete list of expected metadata keys for the specific template. This will not work with more complicated schemas, yet.


avanlinden (approved):

Aryllen:
Windows check still failing due to credentials. I tried a fix, but it's still not working. Will make a separate issue for that part.

R/get-template.R
```diff
@@ -0,0 +1,179 @@
...
+  ## Use synID if given
+  if (!is.na(synID)) {
+    return(get_template_keys_synID(syn = syn, synID = synID, ...))
+  }
+  if (!is.na(id)) {
+    ## Check if id was used to give a synID
+    if (grepl("^syn", id)) {
+      return(get_template_keys_synID(syn = syn, synID = id, ...))
+    }
+    ## If not a synID, assume a schema id
```
kelshmo, line 31: What is the schema id convention?

Aryllen: organization.name-schema.name-version

For example, the schema in the default config is:
https://github.com/Sage-Bionetworks/dccvalidator/blob/8d7b75c88a2f0b7ee76640d2d22ef3fbcd4dacdb/inst/config.yml#L28

This is registered under the organization sysbio.metadataTemplates with the schema name assay.rnaSeq. There is no version, but if there was, it would be something like `sysbio.metadataTemplates-assay.rnaSeq-0.0.1`, similar to versions in the synapseAnnotations repository.

More info can be found in confluence for [Synapse schema syntax](https://sagebionetworks.jira.com/wiki/spaces/PLFM/pages/989167828/Schemas+for+extending+Synapse+Objects#Name-spacing).

kelshmo (approved):
Pulled this locally and played with some of the new functions. Very cool and your tests have good coverage between .csv/.xlsx templates and registered schema templates. 

Unrelated to this PR, when I run your tests locally, I get some failures as `attempt_instantiate` is an unexported function in synapse-helpers.R. I'm curious how this function gets loaded when your actions run tests.

Aryllen:
> Unrelated to this PR, when I run your tests locally, I get some failures as attempt_instantiate is an unexported function in synapse-helpers.R. I'm curious how this function gets loaded when your actions run tests.

That's interesting. Are you loading the whole package?

There's 3 different ways I test things: running individual tests for spot checks, `devtools::check()`, and `devtools::test()`. In all cases there's an initial `devtools::load_all()` call before testing. `devtools::check()` is the only one that you don't have to explicitly tell R to load the package because it's included in the function. If you only import the package (i.e. `library(dccvalidator)`), it would probably give some errors when testing.

PR #498, opened 2021-07-09
Add ability to select biospecimen template by type

### LINKED ISSUE #495: Add distinction between in vivo and in vitro biospecimen templates
The biospecimen template used currently depends on the species. Need to add additional feature for whether the template is for in vitro or in vivo specimens.

Aryllen:
Fixes #495. Depends on #497.

- New config options for having biospecimen templates by type and signaling to the UI to add the feature. The radiobutton options will reflect what the types are called in the template.
- Fix for drosophila logic. When switched to OAuth, lost the conditionalPanel logic for individual metadata. Added it back.
- Update to `gather_template_ids` to accept `biospecimen_type`.
- Update to `get_template_synIDs` to only gather synIDs and unlist the biospecimen templates if nested
- Updated docs
- Updated tests

Questions:

- Is "Specimen Type" a good name for this feature?
- What should the PEC config look like with this new addition?
- Should in vitro biospecimen be an option for all species in AD, except drosophila?

~~I will make a test version of the app on Friday.~~ [Test app here.]( https://www.synapse.org/#!Wiki:syn25976061/ENTITY) Note that this only has two assays as it's the default config.

Reviewing the code itself may be a bit confusing since this relies on a different branch that is also a PR, atm. I'll update this branch once I merge the other PR (hopefully later today).


avanlinden:
> Is "Specimen Type" a good name for this feature?

I'd say maybe go with "Biospecimen Type", to explicitly link it to the metadata template. They will know which biospecimen template we've told them to use so hopefully that will make sense. Open to other opinions though!

Aryllen:
Based on conversation with @avanlinden, updated the name of the feature to "Biospecimen Type". I also updated the default config to show different ways to name the types (change the species to see the different options or how the feature goes away completely for drosophila). Whatever is first in the config list of biospecimen-type templates will be selected by default.

R/df-helpers.R
```diff
@@ -59,10 +65,16 @@ gather_template_id_individual <- function(species) {
 }
 
 ## gather_template_ids helper
-gather_template_id_biospecimen <- function(species) {
+gather_template_id_biospecimen <- function(species, biospecimen_type) {
   templates <- get_golem_config("templates")$biospecimen_templates
```
kelshmo, line 69: Having never worked with golem not sure this is relevant, but should templates be an input parameter and then `get_golem_config` called from the function `gather_template_id_biospecimen`? I'm thinking about how I integrate `config` with function arguments.

R/df-helpers.R
```diff
@@ -38,12 +38,18 @@ get_metadataType_indices <- function(data, meta_types) {
 #' @param species The species needed to specify the correct biospecimen
 #' or individual templates (default `NA`).
 #' @param assay The assay needed to specify the correct assay template.
+#' @param biospecimen_type The type of biospecimen template needed
```
kelshmo, line 41: Are accepted types "in_vitro" or "other"? What will happen if another value is passed?

R/metadata-template-dictionary.R
```diff
@@ -307,8 +307,16 @@ generate_key_description <- function(annots) {
...
     unlist()
+  temps[grepl("^syn", temps)]
 }
+
+## get_template_synIDs helper
+get_template_synIDs_nested <- function(templates) {
+  templates %>%
+    unname() %>%
+    unlist()
+}
```
kelshmo, line 322:
```suggestion
}

```

vignettes/customizing-dccvalidator.Rmd
```diff
@@ -124,11 +124,26 @@ To install the dccvalidator instead of forking the repository:
...
-  an excel file with multiple sheets, the first sheet will be used. For templates accessed via JSON schema id, these should have a `properties` object listing all column names (a.k.a. keys) that should be present in the metadata.
+  use for validation. These may be a Synapse synID or the id of a
+  Synapse-registered JSON schema. For templates accessed via synID, these should
+  be either .xlsx or .csv files, where the column names reflect the required
+  columns in the template. If the template is an excel file with multiple
+  sheets, the first sheet will be used. For templates accessed via JSON schema
+  id, these should have a `properties` object listing all column names (a.k.a.
+  keys) that should be present in the metadata. The `manifest_template` should
+  have only one template. `individual_templates` and `biospecimen_templates`
+  should be broken up by species (the species must be included in the
```
kelshmo, line 140: I suggest revising the language to clarify "broken up by species": 
* `templates` (including `manifest_template`,......
     `individual_templates:`
                `<name species:> templates must be species-specific and the species must be named in a nested list under        `individual_templates`. 

It is difficult to capture the nested/indented relationship in this comment! Maybe instead of being explicit about the nested format, you could link to the appropriate line in the example config. Same thought goes for `biospecimen_templates`.

kelshmo (approved):
Approving, as the test app is working well! I included some comments of where the code/documentation could be made more explicit/clarified.

R/df-helpers.R
Aryllen: This accepts any values from the biospecimen template section of the config. The `biospecimen_type`s that will appear in the UI options will be pulled from the list of named biospecimen templates in the config. This adds extra flexibility for more types of biospecimen templates in the future, and gives you control over how the `biospecimen_type` appears in the app.

R/df-helpers.R
Aryllen: I'm not exactly sure what you mean. `config` and `golem` (`get_golem_config()` calls `config::get()`) tend to assume a single list of options, not nested lists. However, it is a good point that I could grab the list of templates from the config in `gather_template_ids()` and pass it along, instead of grabbing it multiple times. I'll update this to be a little more efficient.

vignettes/customizing-dccvalidator.Rmd
Aryllen: This paragraph seemed cumbersome to me, as well! Thanks for the suggestion on how to make it better. I'll update it.

Aryllen: Updated.

R/df-helpers.R
Aryllen: Updated and made it a little nicer (IMO) for how I'm referencing specific lists of templates.

Aryllen:
Realized that I would need a biospecimenType annotation on the uploaded metadata files in order to make dccmonitor compatible. Updated to add this annotation on biospecimen metadata. Also made a few updates based on @kelshmo's feedback.

Aryllen:
Most of the checks have finished successfully after the latest merge of master. I am expecting the Windows check to fail, but that has to do with the setup of the .synapseConfig, not the package. Merging.

Note: dccmonitor should **not** be updated with the latest version of dccvalidator until I make dccmonitor compatible. Similarly, until dccmonitor is compatible and updated, the biospecimen metadata in dccmonitor will be checked against the main species-based biospecimen metadata; it will not check based on biospecimen type.

R/df-helpers.R
kelshmo: fabulous.

R/df-helpers.R
kelshmo: Your updated code is spot on with my comment. Looks good!

PR #510, opened 2021-08-20
add link to documentation submission form

avanlinden:
Fixes # (NA)

@RYaxley pointed out that we link to the doc submission Synapse form on the "Using the App" page but not actually on the "Study Documentation" page, which caused some confusion about where to actually submit the documentation. I suggest adding a link specifically to the form on the documentation page within the app, to make it clear that we're not submitting via the app any more. Does this look ok?

-


pitviper6 (approved):
It looks good to me!

Aryllen (approved):
Looks good! Double-checked on the server.

PR #517, opened 2021-09-07
Check for non-ascii characters and block validation

### LINKED ISSUE #516: Check for non-ascii characters
Check for non-ascii characters in the files. This has been a big problem with the metadata files, where odd characters have made their way into the data.

Aryllen:
Fixes #516

- Add `check_invalid_characters()`
- Check for invalid characters before uploading data and validating. If there are invalid characters, use additional helper functions to create an error message with the name of the column that the invalid characters were found in

Note: this is not the most graceful of additions. I know this has been a big problem, though, so hopefully this will help.


Aryllen:
Test version here: https://www.synapse.org/#!Wiki:syn25976061/ENTITY

avanlinden:
@Aryllen The test app can't find the function `summarize_all_failed_checks`.

Screen Shot 2021-09-07 at 12 11 27 PM

Aryllen:
> @Aryllen The test app can't find the function `summarize_all_failed_checks`.
> 
> Screen Shot 2021-09-07 at 12 11 27 PM

Ugh. I'll fix that...

R/check-invalid-characters.R
```diff
@@ -0,0 +1,76 @@
...
+#' check_invalid_characters(dat)
+check_invalid_characters <- function(data,
+                             success_msg = "There are no invalid characters",
+                             fail_msg = "There is an invalid character in a column") { #nolint
+  if (is.null(data)) {
+    return(NULL)
+  }
+  has_invalid <- purrr::map_lgl(data, ~ contains_invalid(.))
+  behavior <- glue::glue(
+    "Only standard ascii characters are allowed."
```
kelshmo, line 25: Where does `behavior` get surfaced? I'm not seeing this message, and it would be helpful to know why the characters are invalid.

kelshmo (commented):
Screen Shot 2021-09-07 at 3 20 38 PM
The error message is visualized however, I dropped a comment to clarify where "Only standard ascii characters are allowed." is getting surfaced. It would be clearer if the message returned in the validator tab included this explicitly.

R/check-invalid-characters.R
Aryllen: That behavior isn't surfaced in the current version of the app. Ideally, someone could add this to the `check_all` function instead of having it block upload (which is what I'm using it for in this version), which is why I made it in a similar format to the other types of validation checks. The downside to using it in `check_all` and having the result pop up in the results boxes is that other pieces of the app (summary/details tab and other validation checks) end up breaking, as well.

I added a bit of text to the error summary (it's updated on the test version). Now it should say:
```
Error: Only standard ascii characters are allowed in the files.
There are invalid characters in the manifest columns: study
There are invalid characters in the individual metadata columns: individualID, sex
```

How is that, @kelshmo?

avanlinden: That will definitely be helpful. I will say I appreciate the error message identifying which file and column have non-standard characters!

kelshmo (approved):

